PR #24, opened 2020-01-29
update approval workflows

cbudz:
This PR updates Approval documentation to reflect the change to a group-based approval workflow.

### DIFF
--- a/doc/modules/con_Approval.adoc
+++ b/doc/modules/con_Approval.adoc
@@ -3,7 +3,7 @@
 [id="Approval"]
 = Working with approval
 
-Catalog supports multi-level approval workflows, providing the flexibility to create a tiered administrative process that includes selected organizational groups. Administrators can design approval workflows, view request details, approve or deny requests, and monitor status of existing requests. Once created, an approval workflow can be assigned to Catalog portfolios.
+Catalog supports group-based approval workflows, providing the flexibility that includes selected organizational groups. Administrators can assign approval groups, view request details, approve or deny requests, and monitor status of existing requests. Once created, an approval workflow can be assigned to Catalog portfolios.
 
 == Creating approval workflows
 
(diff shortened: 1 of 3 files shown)


doc/modules/con_Approval.adoc
```diff
@@ -3,7 +3,7 @@
 [id="Approval"]
 = Working with approval
 
-Catalog supports multi-level approval workflows, providing the flexibility to create a tiered administrative process that includes selected organizational groups. Administrators can design approval workflows, view request details, approve or deny requests, and monitor status of existing requests. Once created, an approval workflow can be assigned to Catalog portfolios.
+Catalog supports group-based approval workflows, providing the flexibility that includes selected organizational groups. Administrators can assign approval groups, view request details, approve or deny requests, and monitor status of existing requests. Once created, an approval workflow can be assigned to Catalog portfolios.
```
gmcculloug: Approval workflows can currently be applied to Portfolios, Products and Platform Inventories.  The intention is to support Platforms as well, but is dependent on tagging support on the Sources platform service which is in-progress.

cbudz: Thanks @gmcculloug, I've added this detail to the doc. For the time I've left platforms in there since its in progress.

doc/modules/con_Approval.adoc
```diff
@@ -3,19 +3,21 @@
 [id="Approval"]
 = Working with approval
 
-Catalog supports multi-level approval workflows, providing the flexibility to create a tiered administrative process that includes selected organizational groups. Administrators can design approval workflows, view request details, approve or deny requests, and monitor status of existing requests. Once created, an approval workflow can be assigned to Catalog portfolios.
+Catalog supports group-based approval processes applied to portfolios, products, and platform inventories, providing the flexibility to include selected organizational groups in approving requests.
+
+Approval administrators can create approval groups and processes, view request details, approve or deny requests, and monitor status of existing requests.
```
gmcculloug: We need to include a note that the Approval Administrator also needs to be an Organization Admins on the Platform to create groups and list the groups when creating Approval processes.  

Organization Admins are defined at the SSO level, not in the Platform RBAC service.

cbudz: I've added a note in the lates commit. I think we'll need to add a separate doc explaining SSO level vs RBAC service, but not sure how Catalog specific this would be at this point?

PR #26, opened 2020-02-19
Update edit portfolios

cbudz:
This PR updates the edit portfolio workflow:

1. Editing Portfolios is limited to the current state of edit -> change name and description.
2. Set approval process description is updated and the procedure for applying an approval process to a portfolio is now current.

### DIFF
--- a/doc/modules/con_Assigning_approval_workflow.adoc
+++ b/doc/modules/con_Assigning_approval_workflow.adoc
@@ -1,8 +1,8 @@
 // Module included in the following assemblies:
 // assembly_Administration_guide.adoc
 [id="Assigning_approval_workflow"]
-= Assigning approval workflows to portfolios
+= Setting approval processes for portfolios
 
-Assign approval workflows, once created, to your portfolios. Each approval workflow will apply to all products in a given portfolio.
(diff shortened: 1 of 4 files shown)


doc/modules/con_Assigning_approval_workflow.adoc
```diff
@@ -1,8 +1,8 @@
 // Module included in the following assemblies:
 // assembly_Administration_guide.adoc
 [id="Assigning_approval_workflow"]
-= Assigning approval workflows to portfolios
+= Setting approval processes for portfolios
 
-Assign approval workflows, once created, to your portfolios. Each approval workflow will apply to all products in a given portfolio.
+Govern access to products by setting the approval processes for the portfolios in which they are organized. Once approval processes are set for a portfolio, each product ordered in that portfolio will require approval from the groups defined in the process. Approval processes will apply to all products in a given portfolio.
```
gmcculloug: The last two sentences seem redundant:
1. Once approval processes are set for a portfolio, each product ordered in that portfolio will require approval from the groups defined in the process. 
2. Approval processes will apply to all products in a given portfolio.

I would keep the first one which is more descriptive.  But if you think we should keep both that is fine as well.

Otherwise the rest of the changes in the PR look good.

cbudz: @gmcculloug that is quire redundant. I've removed the last sentence.

PR #28, opened 2020-03-04
Surveys

cbudz:
Adding a few things in this PR:

- Overview of Surveys
- Procedure on how to access surveys from the product detail view
- Product attributes file
- Files for "Working with Products" guide

- Directory structure also revised with this PR.

### DIFF
--- a/downstream/modules/con_Surveys.adoc
+++ b/downstream/modules/con_Surveys.adoc
@@ -23,5 +23,5 @@ Additionally, use the properties editor to hide, disable, or set a fields to rea
 
 [NOTE]
 ====
-Provide an initial value for any required field set to *Hidden* or *Disabled*.
+Provide an initial value for any required field set to *Hidden* or *Disabled*. This includes `read only` fields. 
 ====
--- a/downstream/modules/con_Surveys.adoc
+++ b/downstream/modules/con_Surveys.adoc
@@ -1,0 +1,27 @@
+
+[id="concept-surverys_{context}"]
+
+= {productname} survey editing
+
+When ordering a product in {productname}, users may be required to provide additional information to complete the request. This prompt is created by a survey associated with the product. Initially created and edited in Ansible Tower, Catalog Administrators and portfolio users with *update* permissions can edit the survey in {productname}. Once submitted and validated, the surveys pass user-submitted extra variables to the job template run on Ansible Tower on the execution of an order.
+
+See _Surveys_ in the link:https://docs.ansible.com/ansible-tower/latest/html/userguide/index.html[Ansible Tower User Guide] to learn more creating and editing surveys for job templates.
+
+
+Using the survey properties editor, you can:
+
+* Set *labels*, *helper text* or *placeholders* to enhance the user experience of users providing information
+* Further restrict existing validation parameters.
+* Change validation messages.
+* Hide, disable or set to read-only chosen fields in the survey.
+
+== Survey properties editor
+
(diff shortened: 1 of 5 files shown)


gmcculloug:
@lindgrenj6 @syncrou I am all for adding as much documentation as needed to the rake task, but I am concerned that we are including links to internal URLs (`https://sonarqube.corp.redhat.com`) which are unreachable for community users.

I worked with @lindgrenj6 on running a local docker version of sonar which seems like a better fit here.  Or maybe there is another option we could consider.  We can document how to push results to internal servers elsewhere.  Thoughts?

downstream/modules/con_Surveys.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+== Survey properties editor
+
+Values for surveys are defined in Ansible Tower and the {productname} *properties editor* allows Catalog Administrators to restrict survey options, such as adjusting default values and removing items from drop-down menus.
+
+Additionally, use the properties editor to hide, disable, or set a fields to read only.
+
+[NOTE]
+====
+Provide An initial value for any required field set to *Hidden* or *Disabled*.
```
gmcculloug: This applies to any field type that cannot be modified, `read only` would be included in this list.

downstream/modules/con_Surveys.adoc
```diff
@@ -0,0 +1,27 @@
...
+Using the survey properties editor, you can:
+
+* Set *labels*, *helper text* or *placeholders* to enhance the user experience of users providing information
+* Further restrict existing validation parameters.
+* Change validation messages.
+* Hide, disable or set to read-only chosen fields in the survey.
+
+== Survey properties editor
+
+Values for surveys are defined in Ansible Tower and the {productname} *properties editor* allows Catalog Administrators to restrict survey options, such as adjusting default values and removing items from drop-down menus.
```
gmcculloug: Same `Catalog Administrator` question.  Should it use substitution?

downstream/modules/con_Surveys.adoc
```diff
@@ -0,0 +1,27 @@
+
+[id="concept-surverys_{context}"]
+
+= {productname} survey editing
+
+When ordering a product in {productname}, users may be required to provide additional information to complete the request. This prompt is created by a survey associated with the product. Initially created and edited in Ansible Tower, Catalog Administrators and portfolio users with *update* permissions can edit the survey in {productname}. Once submitted and validated, the surveys pass user-submitted extra variables to the job template run on Ansible Tower on the execution of an order.
```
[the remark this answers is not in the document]
cbudz: I based `Catalog Administrators` on the RBAC role name but let me know if you think it's a better idea to replace this with a variable. If so I'll open a ticket to add roles to the variables file if we expect any changes. 

IMO referring to them both as Ansible Tower Templates seems most reasonable.

downstream/modules/con_Surveys.adoc
cbudz: Please see above. RBAC role here, can implement variable if you think its a good idea.

downstream/modules/con_Surveys.adoc
cbudz: Updated to specify this extends to `read only` fields.

cbudz:
hey @gmcculloug I went ahead and made some updates to the files here. If you think extending variables to include the RBAC roles is a good idea I can add those to the list. Let me know.

downstream/modules/con_Surveys.adoc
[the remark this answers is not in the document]
gmcculloug: Let's leave it alone for now and can address if this changes in the future.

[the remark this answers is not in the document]
cbudz: 👍

PR #29, opened 2020-03-16
New onboarding

cbudz:
This PR adds the text of the onboarding guide


downstream/titles/onboarding/master.adoc
```diff
@@ -63,7 +49,20 @@ To use Automation Service Catalog as a Lighthouse customer, create a new Red Hat
...
+The required entitlements are added and you can now access Automation Service Catalog.
+
+== Groups and users
+
+When configuring Automation Service Catalog for your needs, consider the scope of your user environments and how those users are grouped.
+
+Determine the required level of access for your users based on the following predefined roles in Automation Service Catalog:
+
+* *Organization Administrator*: a Red Hat account user who can manage users and control their access and permissions.
+* *Approval Administrator*: 	An Approval administrator role that grants create, read, update and destroy permissions.
```
mkanoor: @bzwei Should we be adding the objects that are involved here, request, stage, action etc.?

downstream/titles/onboarding/master.adoc
```diff
@@ -63,7 +49,20 @@ To use Automation Service Catalog as a Lighthouse customer, create a new Red Hat
...
+* *Approval Approver*: 	An Approval approver role that grants read and create permissions.
+* *Approval User*: An approval user role which grants permissions to read/create a request.
+* *Catalog Administrator*: A catalog administrator roles grants create, read, update, delete and order permissions.
```
mkanoor: @gmcculloug Should we add permissions relate to the Portfolio object

downstream/titles/onboarding/master.adoc
bzwei: Approval Administrator role grands create, read, update and destroy permissions to approval process objects, plus permissions for approval approver and approval user roles.

downstream/titles/onboarding/master.adoc
```diff
@@ -0,0 +1,151 @@
...
+
+== Groups and users
+
+When configuring Automation Service Catalog for your needs, consider the scope of your user environments and how those users are grouped.
+
+Determine the required level of access for your users based on the following predefined roles in Automation Service Catalog:
+
+* *Organization Administrator*: a Red Hat account user who can manage users and control their access and permissions.
+* *Approval Administrator*: 	An Approval administrator role that grants create, read, update and destroy permissions.
+* *Approval Approver*: 	An Approval approver role that grants read and create permissions.
```
bzwei: Approval approver has read permission to requests, create permission to actions.

cbudz: hey @bzwei and @mkanoor if the Approval approver only has read access to requests I think we need to rename that role as in its present state it suggests the user can approve requests. So if we're creating groups for approval processes, the users in that group all need to be Approval administrators to approval orders? Am I understanding that correctly?

cbudz: @bzwei @mkanoor Approval approver does also have approve/deny permissions, right? That is the primary role for adding to an approval group?

downstream/titles/onboarding/master.adoc
cbudz: Working with UX to provide more descriptive explanations. RBAC docs have full description at present and we will link over to that docs set for any in-depth needs customers may have.

downstream/titles/onboarding/master.adoc
cbudz: @mkanoor directed me to the UX effort for these topics, working with them on providing concise descriptions per role.

mkanoor: @cbudz It might make sense to track this file
https://github.com/RedHatInsights/rbac-config/blob/master/configs/approval.json
and this one for catalog
https://github.com/RedHatInsights/rbac-config/blob/master/configs/catalog.json

It has all the permissions for each of the roles so we can match it with the docs.

cbudz:
Merging this content to pull in new directory structure.

PR #79, opened 2021-06-10
Add install guide

cbudz:
Revamping the install guide PR here


cbudz:
merging these files so updated directories can be used for new work

PR #141, opened 2021-07-20
AAP 396 Remove references to quay.io and replace with registry.redhat.io

cbudz:
Per: https://issues.redhat.com/browse/AAP-396

Also updated sec 4.3 -> step 7 on pushing image to the automation hub container registry.


TarikFD:
Changes look good. Merging

PR #143, opened 2021-07-21
Update YAML examples and play description

samdoran:
Ansible does not make use of multiple YAML documents, so the document separator can be omitted from examples.

Update descriptions of basic playbook components to be more accurate and avoid the use of latin phrases.


samccann (approved):
LGTM

TarikFD:
Thanks @samdoran . This looks good.

Merging

samdoran:
Thanks!

PR #154, opened 2021-08-09
First draft of operator install procedures

TarikFD:
First draft of the platform operator installation procedures on OpenShift. I edited some of the assemblies that were created previously to reflect the TLS secrets and Ingress configurations in the form view. This draft encompasses both hub and controller installations.

Preview (VPN required to view): http://file.rdu.redhat.com/kevchin/operator/operator-procs/tmp/en-US/html-single/

Note: These drafts only reflect some of the installation scenarios we were tasked to do. Once we get a clearer understanding on how the procedures change on a external/local database, we can add those later.

### DIFF
--- a/downstream/assemblies/platform/assembly-installing-controller-operator-external-db.adoc
+++ b/downstream/assemblies/platform/assembly-installing-controller-operator-external-db.adoc
@@ -21,32 +21,36 @@ You can use these instructions to install the {ControllerName} operator on {OCP}
 
 == Prerequisites
 
-* You have installed the catalog containing the {PlatformName} operators
+* You have installed the :PlatformName: catalog in Operator Hub.
 
 == Installing the {ControllerName} operator
 
 . Log in to {OCP}.
 . Navigate to *Operators* -> *Operator Hub*.
-. Locate the {PlatformName} catalog and click *Create Instance*.
-
+. Locate the {PlatformName} operator and click *Create Instance*.
++
 [NOTE]
 ====
 {OCP} clusters running on AWS do not support ReadWriteMany without adding NFS or other storage.
 ====
++
+. Upon successful installation of the platform operator, click on the *View operator* button on the installation confirmation message.
+.. You may also find the installed operator by navigating to *Operators* -> *Installed Operators*
+. Locate the {ControllerName} operator and click *Create instance*.
 
 .Optional - Configure your {ControllerName} operator route options
 . Click *Advanced configuration*.
-. Under “Tower Ingress type”, click the dropdown menu and select *Route*.
-. Under “Route DNS host”, enter a common host name that the route answers to.
-. Under “Route TLS termination mechanism”, click the dropdown menu and select *Edge* or *Passthrough*.
-. Under “Route TLS credential secret”, click the dropdown menu and select a secret from the list.
+. Under *Tower Ingress type*, click the drop-down menu and select *Route*.
+. Under *Route DNS host*, enter a common host name that the route answers to.
+. Under *Route TLS termination mechanism*, click the drop-down menu and select *Edge* or *Passthrough*.
+. Under *Route TLS credential secret*, click the drop-down menu and select a secret from the list.
 
 
 .Optional - Configure the Ingress type for your {ControllerName} operator
 . Click *Advanced Configuration*.
-. Under “Tower Ingress type”, click the dropdown menu and select *Ingress*.
-. Under “Tower Ingress annotations”, enter any annotations to add to the ingress.
-. Under “Tower Ingress TLS secret”, click the dropdown menu and select a secret from the list.
+. Under *Tower Ingress type*, click the drop-down menu and select *Ingress*.
+. Under *Tower Ingress annotations*, enter any annotations to add to the ingress.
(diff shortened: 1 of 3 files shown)


downstream/assemblies/platform/assembly-installing-controller-operator-external-db.adoc
```diff
@@ -8,48 +8,49 @@ ifdef::context[:parent-context: {context}]
...
-= Installing {ControllerName} on {OCP} with an external database and routed web traffic
+= Installing {ControllerName} on {OCP} with an external database
 
 
 :context: installing-contr-operator-external
 
 
 [role="_abstract"]
-You can use these instructions to install the {HubName} operator on {OCP} with an external PostgreSQL 12 database and route, as well as specify customer resources.
+You can use these instructions to install the {ControllerName} operator on {OCP} with an external PostgreSQL 12 database and route, as well as specify customer resources.
```
ariordan-redhat, line 18: Do you think we should keep using **{ProductName} operator**, or the attributes defined specifically for the operator? 
**{ProductName} operator** is cleaner. 
The only advantage to using the defined operator names is a possible future name change for the operators.

downstream/assemblies/platform/assembly-installing-controller-operator-external-db.adoc
```diff
@@ -8,48 +8,49 @@ ifdef::context[:parent-context: {context}]
...
 
 // mirrors AWX operator flow
 
 == Prerequisites
 
 * You have installed the catalog containing the {PlatformName} operators
```
ariordan-redhat, line 24: Should this be "You have installed the {PlatformName} operator" ?

Also, it might be an idea to break the procedural content in the assemblies out into separate procedure and concept modules.

downstream/assemblies/platform/assembly-installing-controller-operator-external-db.adoc
```diff
@@ -8,48 +8,49 @@ ifdef::context[:parent-context: {context}]
...
 
-.Installing the {HubName} operator
+== Installing the {ControllerName} operator
 
 . Log in to {OCP}.
 . Navigate to *Operators* -> *Operator Hub*.
-. Locate the {PlatformName} catalog and click on it.
```
ariordan-redhat, line 30: Should this be "Locate the {PlatformName} operator " ?

downstream/assemblies/platform/assembly-installing-controller-operator-external-db.adoc
```diff
@@ -8,48 +8,49 @@ ifdef::context[:parent-context: {context}]
...
-. Click *Install*
```
ariordan-redhat, line 31: It might be useful here to say that upon successful installation of Platform Operator, a message appears with a "view Operator" button. Click this button to view the resources provided by the Platform Operator (Hub and Controller deployment and management)

downstream/assemblies/platform/assembly-installing-hub-operator-external-db.adoc
```diff
@@ -23,35 +23,51 @@ You can use these instructions to install the {HubName} operator on {OCP} with a
...
+. Click *Advanced configuration*.
+. Under “Ingress type”, click the dropdown menu and select *Route*.
+. Under “Route DNS host”, enter a common host name that the route answers to.
+. Under “Route TLS termination mechanism”, click the dropdown menu and select *Edge* or *Passthrough*.
+. Under “Route TLS credential secret”, click the dropdown menu and select a secret from the list.
+
+
+.Optional - Configure the Ingress type for your {HubName} operator
+. Click *Advanced Configuration*.
+. Under “Ingress type”, click the dropdown menu and select *Ingress*.
```
ariordan-redhat, line 51: Should the GUI text be bold  - *Ingress type* ? 
https://redhat-documentation.github.io/asciidoc-markup-conventions/ - see "GUI Text"

Not sure if it would be accurate to say something like 
In the *Ingress Type* menu, select *Ingress* 
In the *Ingress annotations* field, enter any annotations to add to the ingress.

ariordan-redhat (commented):
Just a couple of suggestions.

downstream/assemblies/platform/assembly-installing-hub-operator-external-db.adoc
TarikFD: You're right, it should be bold. Says the same thing on page 208 on the IBM style guide.

I've also changed dropdown to drop-down, as I found from glancing at the style guide

downstream/assemblies/platform/assembly-installing-controller-operator-external-db.adoc
TarikFD: I'm honestly not sure. I see it working either way.

@cbudz what do you think? We can either go with {ProductName} Operator or create its own {ProductOperator}

cbudz: @TarikFD lets go with {PlatformOperator} just in case they change it.

ariordan-redhat (approved):
Looks good to me

PR #160, opened 2021-08-12
update product number in docinfo.xml

cbudz:
Quick fix here so I can set-up preview for 2.0-ea docs set.


ariordan-redhat (approved):
Looks good to me

PR #224, opened 2022-01-12
bugfix: Asciidoc artefact visible in rendered Automation Mesh guide

ariordan-redhat:
**Problem:**

At the end of the preface in the [Automation Mesh Guide](https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.1/html/red_hat_ansible_automation_platform_automation_mesh_guide/pr01), an artefact from Asciidoc (`:leveloffset:+1`) is visible in the published docs.

We think this is caused by a hidden character in the blank line between the preface and the first included assembly.

**Solution:**
Deleted the blank line and added in another one.
Also added a blank line at the end of of the doc, as Git didn't detect a diff when the blank line that was causing the problem was replaced.

Before:
image

After:
image


TarikFD:
Looks good!

PR #239, opened 2022-02-09
AAP-1637 Added note about enabling Simple Content Access

TarikFD:
Jira request: https://issues.redhat.com/browse/AAP-1637

Added a note in the aap installer > subscriptions section regarding enabling simple content access, some background info, and a link to the procedures on enabling it. 

Titles affect: `titles/aap-installation-guide`


rogrange (commented):
The note is clear in its explanation. So I would say this is fine for now. However, I would like to note the front paragraph is confusing, and I think it could do with a rewrite, or at the very least, a good edit.

TarikFD:
> The note is clear in its explanation. So I would say this is fine for now. However, I would like to note the front paragraph is confusing, and I think it could do with a rewrite, or at the very least, a good edit.

Thanks. Funny enough we had a ticket about editing that first paragraph [here](https://issues.redhat.com/browse/AAP-1731). I drafted a PR for that section [here](https://github.com/RedHatInsights/red-hat-ansible-automation-platform-documentation/pull/242) . Let me know if you think that works or if we could work on it more

PR #242, opened 2022-02-11
AAP-1731 Clarifying subscriptions requirements for installing AAP

TarikFD:
Jira: https://issues.redhat.com/browse/AAP-1731

Requirements have been updating so that all nodes require a red hat subscription for the AAP installer to work. Edited the concept module to clarify this point, and added a troubleshooting step at the end as per Satoe's notes.

See section 1.3 of the preview below
Preview link (VPN required): http://file.rdu.redhat.com/kevchin/aap_sub/tmp/en-US/html-single/#proc-attaching-subscriptions_planning

Titles affected: `titles/aap-installation-guide`

### DIFF
--- a/downstream/modules/platform/proc-attaching-subscriptions.adoc
+++ b/downstream/modules/platform/proc-attaching-subscriptions.adoc
@@ -5,10 +5,9 @@
 = Attaching your {PlatformName} subscription
 
 [role="_abstract"]
-You *must* have valid subscriptions attached before installing {PlatformName}. Attaching an {PlatformNameShort} subscription enables {HubName} repositories.
+You *must* have valid subscriptions attached on all nodes before installing {PlatformName}. Attaching your {PlatformNameShort} subscription allows you to access subcription-only resources, such as {HubName} repositories.
 
-A valid subscription needs to be attached to the {HubName} node only. Other nodes do not need to have a valid subscription attached, even if the `[automationhub]` group is blank, given this is done at the `repos_el` role level and that this role is run on both `[automationcontroller] and `[automationhub]` hosts.
-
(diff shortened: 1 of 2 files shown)


rogrange:
I think the first paragraph is fine, however, is not the second paragraph redundant? You have the exact steps following, and I am at a loss to see what use duplicating those steps would serve. I believe we can delete this paragraph and still maintain cohesion and clarity. Thoughts?

"To attach subscriptions to all your Ansible Automation Platform nodes, start by obtaining the pool_id of your Red Hat subscription, then run the subscription-manager attach command."

TarikFD:
> 

Good catch, I agree with your point. Let me remove the second paragraph

downstream/modules/platform/proc-attaching-subscriptions.adoc
```diff
@@ -5,10 +5,7 @@
 = Attaching your {PlatformName} subscription
 
 [role="_abstract"]
-You *must* have valid subscriptions attached before installing {PlatformName}. Attaching an {PlatformNameShort} subscription enables {HubName} repositories.
-
-A valid subscription needs to be attached to the {HubName} node only. Other nodes do not need to have a valid subscription attached, even if the `[automationhub]` group is blank, given this is done at the `repos_el` role level and that this role is run on both `[automationcontroller] and `[automationhub]` hosts.
-
+You *must* have valid subscriptions attached on all nodes before installing {PlatformName}. Attaching your {PlatformNameShort} subscription allows you to access subcription-only resources, such as {HubName} repositories.
```
simaishi, line 8: There is no longer "automation hub repository" and we only have a repo for AAP.

Also, I'm not sure what the instruction has been, when using bundle installer in the disconnected envionment. But it looks like that part hasn't changed in this PR, so I think we're good?  (What we provide in bundle is what you'd get from AAP repo and customers still need to configure in a way they get RHEL rpms from somewhere).

PR #253, opened 2022-02-28
AAP-1799 Update podman push command when pushing registry images

TarikFD:
Jira issue: https://issues.redhat.com/browse/AAP-1799

When dealing with images from registry.redhat.io , users need to include a `--remove-signatures` flag in order for the `podman push` command to work. Updated the command on the docs plus a note briefly explaining the need for the flag.

See section 3.4 in the preview below

Preview (VPN required): http://file.rdu.redhat.com/kevchin/manage-containers/tmp/en-US/html-single/#push-containers

### DIFF
--- a/downstream/modules/automation-hub/proc-push-container.adoc
+++ b/downstream/modules/automation-hub/proc-push-container.adoc
@@ -25,8 +25,10 @@ $ podman login -u=[username] -p=[password] [automation-hub-url]
 . Push your container image to your {HubName} container registry:
 +
 -----
-$ podman push [automation-hub-url]/[container image name]
+$ podman push [automation-hub-url]/[container image name] --remove-signatures
 -----
++
+NOTE: The `--remove-signatures` flag is required when pushing registry.redhat.com images, as it prevents image recompression issues that causes the command to fail. 
 
 
 .Verification
--- a/downstream/modules/automation-hub/proc-push-container.adoc
+++ b/downstream/modules/automation-hub/proc-push-container.adoc
@@ -28,8 +28,7 @@ $ podman login -u=[username] -p=[password] [automation-hub-url]
 $ podman push [automation-hub-url]/[container image name] --remove-signatures
 -----
 +
-NOTE: The `--remove-signatures` flag is required when pushing registry.redhat.io images, as it prevents image recompression issues that causes the command to fail. 
-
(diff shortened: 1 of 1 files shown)


downstream/modules/automation-hub/proc-push-container.adoc
```diff
@@ -25,8 +25,10 @@ $ podman login -u=[username] -p=[password] [automation-hub-url]
 . Push your container image to your {HubName} container registry:
 +
 -----
-$ podman push [automation-hub-url]/[container image name]
+$ podman push [automation-hub-url]/[container image name] --remove-signatures
 -----
++
+NOTE: The `--remove-signatures` flag is required when pushing registry.redhat.io images, as it prevents image recompression issues that causes the command to fail.
```
ipanova, line 31: NOTE: The `--remove-signatures` flag is required when signed images from registry.redhat.io are pushed to the {HubName} container registry. Push operation re-compresses image layers during the upload. The re-compression is not at all guaranteed to be reproducible and  it is client implementation dependent. This can lead to the image layer digest change and  failed push operation with the error `Error: Copying this image requires changing layer representation, which is not possible (image is signed or the destination specifies a digest)`.

downstream/modules/automation-hub/proc-push-container.adoc
```diff
@@ -25,9 +25,10 @@ $ podman login -u=[username] -p=[password] [automation-hub-url]
 . Push your container image to your {HubName} container registry:
 +
 -----
-$ podman push [automation-hub-url]/[container image name]
+$ podman push [automation-hub-url]/[container image name] --remove-signatures
 -----
-
++
+NOTE: The `--remove-signatures` flag is required when signed images from registry.redhat.io are pushed to the {HubName} container registry. The `push` operation re-compresses image layers during the upload, which is not guaranteed to be reproducible and is client implementation dependent. The `--remove-signatures` flag prevents possible image-layer digest changes during the `push` operation that will cause the operation to fail.
```
ipanova, line 31: The last sentence conceptually is not correct, unfortunately. The flag does not prevent image layers digest change. If it would be up to me, I'd include the error message as I was initially suggesting. It might happen, the customer would not familiarize themselves with docs ahead of push operation and get this error. Seeing the error in docs can save from customer cases :)

ipanova (approved):
looks great!

ariordan-redhat:
Backported to 2.1 release branch ([PR-256](https://github.com/RedHatInsights/red-hat-ansible-automation-platform-documentation/pull/256))

PR #272, opened 2022-03-11
aap-2389 Add delete user module to controller

ariordan-redhat:
Add instructions to delete user accounts from controller so that customers know how to delete users' private information.

The module currently is not included in any doc title.
Adding it to the controller module directory as it's a Privacy Information requirement for Ansible on Azure. The content will appear in the controller docs user guide on docs.ansible.com, but we may need to have the information in the downstream repo.

Jira: [AAP-2389](https://issues.redhat.com/browse/AAP-2389)

### DIFF
--- a/downstream/modules/controller/proc-controller-delete-user.adoc
+++ b/downstream/modules/controller/proc-controller-delete-user.adoc
@@ -1,0 +1,18 @@
+// Module included in the following assemblies:
+[id="proc-delete-user"]
+
+= Deleting a user from {ControllerName}
+
+When you delete a user account, the name and email of the user are permanently removed from {ControllerName}.
+
+.Prerequisites
+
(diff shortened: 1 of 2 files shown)


downstream/modules/controller/proc-controller-delete-user.adoc
```diff
@@ -0,0 +1,18 @@
+// Module included in the following assemblies:
+[id="proc-delete-user"]
+
+= Deleting a user from {ControllerName}
+
+When you delete a user account, the name and email of the user are permanently removed from {ControllerName}.
```
lmalivert, line 6: Suggestion: Replace the word `are` with `is`.  
Example: `are permanently removed` -> `is permanently removed`

lmalivert:
Hi @ariordan-redhat. Aside from the two comments, LGTM 👍🏽

PR #297, opened 2022-04-12
Aap 2055

ariordan-redhat:
Jira: [AAP-2055](https://issues.redhat.com/browse/AAP-2055)

Affects the following title:
titles/aap-operator-installation

Add info about minimum OCP version for installing AAP 2.1 using the operator.


TarikFD (approved):
I agree with this resolution, and the changes build correctly for me. LGTM!

PR #306, opened 2022-05-03
Added note stating user EE container images will not be supported by …

rogrange:
…Red Hat.


TarikFD (approved):
A more neutral sentence structure to consider: 

> Red Hat currently does not support users who choose to provide their own container images when building automation execution environments.

Just my own personal preference, and I think it works either way. Whatever you decide, LGTM!

rogrange:
Yours flows better I think. So how would I go about changing that? I have already pushed it up, and since I have not done this in a while and I do not want to mess up the workflow.

PR #332, opened 2022-05-27
Aap 2007 AAP 2.1 on OCP Pre-Flight Check Documentation

jself-sudoku:
Added new conceptual topic regarding the requirement for `ReadWriteMany` for automation hub when installing AAP Operator on OpenShift Container Platform: 

- **4.2.1. Storage options for Ansible Automation Platform Operator installation on Red Hat OpenShift Container Platform** (new concept module)
    - **Provisioning OCP storage with ReadWriteMany access mode** (new procedure module)

### DIFF
--- a/downstream/assemblies/platform/assembly-installing-hub-operator.adoc
+++ b/downstream/assemblies/platform/assembly-installing-hub-operator.adoc
@@ -27,6 +27,7 @@ You can use these instructions to install the {HubName} operator on {OCP}, speci
 . Locate the *Automation hub* entry, then click *Create instance*.
 
 include::platform/con_storage-options-for-operator-installation-on-ocp.adoc[leveloffset=+2]
+include::platform/proc_provision-ocp-storage-with-readwritemany-access-mode.adoc[leveloffset=+3]
 include::platform/proc-hub-route-options.adoc[leveloffset=+2]
 include::platform/proc-hub-ingress-options.adoc[leveloffset=+2]
 
--- a/downstream/assemblies/platform/assembly-installing-hub-operator.adoc
+++ b/downstream/assemblies/platform/assembly-installing-hub-operator.adoc
@@ -26,8 +26,8 @@ You can use these instructions to install the {HubName} operator on {OCP}, speci
 . Navigate to *Operators* > *Installed Operators*.
 . Locate the *Automation hub* entry, then click *Create instance*.
 
-include::platform/con_storage-options-for-operator-installation-on-ocp.adoc[leveloffset=+2]
-include::platform/proc_provision-ocp-storage-with-readwritemany-access-mode.adoc[leveloffset=+3]
+include::platform/con-storage-options-for-operator-installation-on-ocp.adoc[leveloffset=+2]
+include::platform/proc-provision-ocp-storage-with-readwritemany.adoc[leveloffset=+3]
 include::platform/proc-hub-route-options.adoc[leveloffset=+2]
 include::platform/proc-hub-ingress-options.adoc[leveloffset=+2]
 
--- a/downstream/assemblies/platform/assembly-installing-hub-operator.adoc
+++ b/downstream/assemblies/platform/assembly-installing-hub-operator.adoc
@@ -20,16 +20,14 @@ You can use these instructions to install the {HubName} operator on {OCP}, speci
 == Prerequisites
 
 * You have installed the {PlatformName} operator in Operator Hub.
-* You have
 
 == Installing the {HubName} operator
 
 . Navigate to *Operators* > *Installed Operators*.
 . Locate the *Automation hub* entry, then click *Create instance*.
 
-include::platform/proc-hub-route-options.adoc[leveloffset=+2]
 include::platform/con_storage-options-for-operator-installation-on-ocp.adoc[leveloffset=+2]
-include::platform/proc_provision-ocp-storage-with-readwritemany-access-mode.adoc[leveloffset=+3]
+include::platform/proc-hub-route-options.adoc[leveloffset=+2]
 include::platform/proc-hub-ingress-options.adoc[leveloffset=+2]
 
 Once you have configured your {HubName} operator, click *Create* at the bottom of the form view. {OCP} will now create the pods. This may take a few minutes.
--- a/downstream/assemblies/platform/assembly-installing-hub-operator.adoc
+++ b/downstream/assemblies/platform/assembly-installing-hub-operator.adoc
@@ -20,6 +20,7 @@ You can use these instructions to install the {HubName} operator on {OCP}, speci
 == Prerequisites
 
 * You have installed the {PlatformName} operator in Operator Hub.
+* You have
 
 == Installing the {HubName} operator
 
@@ -27,6 +28,8 @@ You can use these instructions to install the {HubName} operator on {OCP}, speci
 . Locate the *Automation hub* entry, then click *Create instance*.
 
 include::platform/proc-hub-route-options.adoc[leveloffset=+2]
+include::platform/con_storage-options-for-operator-installation-on-ocp.adoc[leveloffset=+2]
+include::platform/proc_provision-ocp-storage-with-readwritemany-access-mode.adoc[leveloffset=+3]
 include::platform/proc-hub-ingress-options.adoc[leveloffset=+2]
 
 Once you have configured your {HubName} operator, click *Create* at the bottom of the form view. {OCP} will now create the pods. This may take a few minutes.
--- a/downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
+++ b/downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
@@ -1,13 +1,14 @@
 [id="con_storage-options-for-operator-installation-on-ocp_{context}"]
 = Storage options for {OperatorPlatform} installation on {OCP}
 
-If your installation scenario includes {HubName}, you must ensure that your storage option for {OperatorPlatform} has been set to `ReadWriteMany`.
+If you are using file-based storage and your installation scenario includes {HubName}, you must ensure that your storage option for {OperatorPlatform} has been set to `ReadWriteMany`.
 
-{HubNameStart} requires `ReadWriteMany` storage for operation or Azure Blob storage or an S3-compliant storage because multiple pods responsible for upload and download of content (for example, collections) need access to the same content. Otherwise, multiple pods are not guaranteed to have access to the same permanent virtual circuit (PVC)  unless it is `ReadWriteMany`.
+{HubNameStart} requires `ReadWriteMany` storage for operation or Azure Blob storage or an S3-compliant storage because multiple pods responsible for upload and download of content (for example, collections) need access to the same content. Otherwise, multiple pods are not guaranteed to have access to the same persistent volume claim (PVC)  unless it is `ReadWriteMany`.
 
-In addition, OpenShift Data Foundation provides either or both a ReadWriteMany or s3-compliant implementation. Also, an NFS storage configuration can be setup to support `ReadWriteMany` and is often the simplest approach. This however, introduces the NFS server as a potential single point of failure.
+In addition, OpenShift Data Foundation provides a `ReadWriteMany` or s3-compliant implementation. Also, an NFS storage configuration can be setup to support `ReadWriteMany` and is often the simplest approach. This, however, introduces the NFS server as a potential single point of failure.
 
 [role="_additional-resources"]
 .Additional resources
 
+* link:https://docs.openshift.com/container-platform/4.10/storage/persistent_storage/persistent-storage-nfs.html[Persistent storage using NFS]
 * IBM's link:https://www.ibm.com/support/pages/how-do-i-create-storage-class-nfs-dynamic-storage-provisioning-openshift-environment[How do I create a storage class for NFS dynamic storage provisioning in an OpenShift environment?]
--- a/downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
+++ b/downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
@@ -3,9 +3,9 @@
 
 If you are using file-based storage and your installation scenario includes {HubName}, ensure that you change the `ReadWriteOnce` default storage option for {OperatorPlatform} to `ReadWriteMany`.
 
-{HubNameStart} requires `ReadWriteMany` storage, Azure Blob storage, or an s3-compliant storage for operation because multiple pods responsible for upload and download of content (for example, collections) need access to the same content. Multiple pods can have access to the same persistent volume claim (PVC) as long as you select the `ReadWriteMany` storage option.
+{HubNameStart} requires `ReadWriteMany` file-based storage, Azure Blob storage, or Amazon s3-compliant storage for operation so that multiple pods can access shared content, such as collections.
 
-In addition, OpenShift Data Foundation provides a `ReadWriteMany` or S3-compliant implementation. Also, an NFS storage configuration can be setup to support `ReadWriteMany` and is often the simplest approach. This, however, introduces the NFS server as a potential single point of failure.
+In addition, {ODFShort} provides a `ReadWriteMany` or S3-compliant implementation. Also, you can set up NFS storage configuration to support `ReadWriteMany`. This, however, introduces the NFS server as a potential, single point of failure.
 
 [role="_additional-resources"]
 .Additional resources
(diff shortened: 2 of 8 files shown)


downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,13 @@
+[id="con_storage-options-for-operator-installation-on-ocp_{context}"]
+= Storage options for {OperatorPlatform} installation on {OCP}
+
+If your installation scenario includes {HubName}, you must ensure that your storage option for {OperatorPlatform} has been set to `ReadWriteMany`.
```
chambridge, line 4: This is true if/when the customer is using "file" based storage. If they selected S3 or Azure than this access mode doesn't mean anything.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,13 @@
...
+
+{HubNameStart} requires `ReadWriteMany` storage for operation or Azure Blob storage or an S3-compliant storage because multiple pods responsible for upload and download of content (for example, collections) need access to the same content. Otherwise, multiple pods are not guaranteed to have access to the same permanent virtual circuit (PVC)  unless it is `ReadWriteMany`.
```
chambridge, line 6: PVC here is "persistent volume claim" not "permanent virtual circuit".

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,13 @@
...
+
+In addition, OpenShift Data Foundation provides either or both a ReadWriteMany or s3-compliant implementation. Also, an NFS storage configuration can be setup to support `ReadWriteMany` and is often the simplest approach. This however, introduces the NFS server as a potential single point of failure.
+
+[role="_additional-resources"]
+.Additional resources
+
+* IBM's link:https://www.ibm.com/support/pages/how-do-i-create-storage-class-nfs-dynamic-storage-provisioning-openshift-environment[How do I create a storage class for NFS dynamic storage provisioning in an OpenShift environment?]
```
chambridge, line 13: We should probably promote a Red Hat link:
https://docs.openshift.com/container-platform/4.10/storage/persistent_storage/persistent-storage-nfs.html

I think the IBM link is a better walk-thru but I don't know if we typically reference docs outside of Red Hat for how to configure OpenShift.

chambridge (changes requested):
Some suggestions and corrections.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
jself-sudoku: Thanks for the feedback here. Was unsure about the technical accuracy. Will propose a new statement.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
jself-sudoku: Thank you for the correction.

jself-sudoku:
Thanks, Chris!

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,14 @@
+[id="con_storage-options-for-operator-installation-on-ocp_{context}"]
+= Storage options for {OperatorPlatform} installation on {OCP}
+
+If you are using file-based storage and your installation scenario includes {HubName}, you must ensure that your storage option for {OperatorPlatform} has been set to `ReadWriteMany`.
+
+{HubNameStart} requires `ReadWriteMany` storage for operation or Azure Blob storage or an S3-compliant storage because multiple pods responsible for upload and download of content (for example, collections) need access to the same content. Otherwise, multiple pods are not guaranteed to have access to the same persistent volume claim (PVC)  unless it is `ReadWriteMany`.
+
+In addition, OpenShift Data Foundation provides a `ReadWriteMany` or s3-compliant implementation. Also, an NFS storage configuration can be setup to support `ReadWriteMany` and is often the simplest approach. This, however, introduces the NFS server as a potential single point of failure.
```
ariordan-redhat, line 8: Is S3 supposed to be uppercase or lowercase?

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,14 @@
+[id="con_storage-options-for-operator-installation-on-ocp_{context}"]
+= Storage options for {OperatorPlatform} installation on {OCP}
+
+If you are using file-based storage and your installation scenario includes {HubName}, you must ensure that your storage option for {OperatorPlatform} has been set to `ReadWriteMany`.
```
ariordan-redhat, line 4:
```suggestion
If you are using file-based storage and your installation scenario includes {HubName}, ensure that the storage option for {OperatorPlatform} is set to `ReadWriteMany`.
```

ariordan-redhat: Out of scope for this PR, I know, but I'm curious whether there's a default setting for the storage option.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+{HubNameStart} requires `ReadWriteMany` storage for operation or Azure Blob storage or an S3-compliant storage because multiple pods responsible for upload and download of content (for example, collections) need access to the same content. Otherwise, multiple pods are not guaranteed to have access to the same persistent volume claim (PVC)  unless it is `ReadWriteMany`.
```
ariordan-redhat, line 6:
```suggestion
{HubNameStart} requires `ReadWriteMany` storage or Azure Blob storage or an S3-compliant storage for operation because multiple pods responsible for upload and download of content (for example, collections) need access to the same content. Multiple pods are guaranteed to have access to the same persistent volume claim (PVC) only if `ReadWriteMany` storage is enabled.
```

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+In addition, OpenShift Data Foundation provides a `ReadWriteMany` or s3-compliant implementation. Also, an NFS storage configuration can be setup to support `ReadWriteMany` and is often the simplest approach. This, however, introduces the NFS server as a potential single point of failure.
+
+[role="_additional-resources"]
+.Additional resources
+
+* link:https://docs.openshift.com/container-platform/4.10/storage/persistent_storage/persistent-storage-nfs.html[Persistent storage using NFS]
```
ariordan-redhat, line 13:
```suggestion
* link:https://docs.openshift.com/container-platform/{OCPLatest}/storage/persistent_storage/persistent-storage-nfs.html[Persistent storage using NFS] in the {OCPShort} _Storage_ guide
```

Just a suggestion: indicate that the link to the instructions for NFS are in the OpenShift docs

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
ariordan-redhat: Do you think it would be useful to have an attribute for OpenShift Data Foundation?

ariordan-redhat (commented):
Some suggestions

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
ariordan-redhat: @jself-sudoku It's not essential, but it might be a good idea in this case, because both option names begin with `ReadWrite`.

jself-sudoku: Thanks, @ariordan-redhat. I have updated the intro sentence to show the default setting.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
jself-sudoku: I actually considered that. I have added two entries to the attributes file for this: 

:ODF: Red Hat OpenShift Data Foundation
:ODFShort: OpenShift Data Foundation

downstream/assemblies/platform/assembly-installing-hub-operator.adoc
```diff
@@ -26,6 +26,8 @@ You can use these instructions to install the {HubName} operator on {OCP}, speci
 . Navigate to *Operators* > *Installed Operators*.
 . Locate the *Automation hub* entry, then click *Create instance*.
 
+include::platform/con_storage-options-for-operator-installation-on-ocp.adoc[leveloffset=+2]
+include::platform/proc_provision-ocp-storage-with-readwritemany-access-mode.adoc[leveloffset=+3]
```
oraNod, line 30: I like that you use underscore between the `con`, `proc`, `ref` prefix in the filenames. I'd honestly like to converge on this as standard in AAP docs (it would help with consistency).

One comment I do have here is that perhaps we can shorten the filenames by omitting things like prepositions and obvious stuff. When filenames get too long it can cause the include directive to span multiple lines, which looks yucky in an editor. Plus long filenames can actually lead to system issues because the full path, including the filename, is taken into account for the maximum number of characters.

How about something like "con_operator-storage-requirements.adoc" and "proc_provisioning-readwritemany-access-mode.adoc"?

It is good to see that the filenames closely match the titles, which is great for navigating the file system, so I'll let you work with it and figure out the best filenames. But I would recommend trimming them down a little. Cheers.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,14 @@
+[id="con_storage-options-for-operator-installation-on-ocp_{context}"]
```
oraNod, line 1: I think we should remove the prefix from the ID for SEO purposes. The ID forms part of the URL and when that includes `con_` etc it is less meaningful. Basically it's just noise to the search engine.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,14 @@
...
+= Storage options for {OperatorPlatform} installation on {OCP}
```
oraNod, line 2: s/options/requirements 

I don't think RWX storage is optional for the hub.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+If you are using file-based storage and your installation scenario includes {HubName}, ensure that you change the `ReadWriteOnce` default storage option for {OperatorPlatform} to `ReadWriteMany`.
```
oraNod, line 4:
```suggestion
If your installation includes {HubName}, you must provision a persistent volume claim with `ReadWriteOnce` access.
```

oraNod: something along those lines imho

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+{HubNameStart} requires `ReadWriteMany` storage, Azure Blob storage, or an s3-compliant storage for operation because multiple pods responsible for upload and download of content (for example, collections) need access to the same content. Multiple pods can have access to the same persistent volume claim (PVC) as long as you select the `ReadWriteMany` storage option.
```
oraNod, line 6:
```suggestion
{HubNameStart} requires `ReadWriteMany` storage, Azure Blob storage, or an s3-compliant storage for operation so that multiple pods can access shared content, such as collections.
```

oraNod: I don't think the 2nd sentence is necessary for minimalism. You're basically explaining RWX access, which is covered in k8s and OCP docs.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+In addition, OpenShift Data Foundation provides a `ReadWriteMany` or S3-compliant implementation. Also, an NFS storage configuration can be setup to support `ReadWriteMany` and is often the simplest approach. This, however, introduces the NFS server as a potential single point of failure.
```
oraNod, line 8: Why do we mention Data Foundation here? Should we use the variable instead `{ODFShort}`?

oraNod: "simplest approach" compared to what? I'm not sure what this is telling the reader to do with storage. Can you use ODF as the underlying storage mechanism?

There's also passive voice in the second sentence "can be setup". That should be a phrasal verb rather than noun form as well. s/setup/set up

oraNod (changes requested):

downstream/assemblies/platform/assembly-installing-hub-operator.adoc
jself-sudoku: Hi Don. I use New Doc to create all of my modularization. So it adds the underscore after the prefix, which brings up a point I meant to ask the team. I noticed a lot of our files do not follow that standard. So I needed to ask about that.

As for shortening the files and leaving out prepositions or other miscellaneous/less important words, unfortunately, New Doc takes that are added within quotes and creates the file names. It doesn't leave out any words. I do like the idea of shortening the file names. So perhaps on my next use of New Doc, I can remove the extraneous words from the title for the purpose of creating the file name and when I edit the content, I can add whatever is necessary for the title. Thanks for helping me come up with an idea to make file names more concise.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
jself-sudoku: I admit I have a technological gap here with this product. Are you referring to the title for this? The SME, Chris Hambridge, reviewed this a couple of weeks ago and did not mention the title being erroneous. That said, it doesn't mean there might be a better way of describing the content. I'll have to circle back to check on this, but he didn't mention any issues with the title. Going with "stet" with this one for now.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
jself-sudoku: Actually, the goal with this is to provision with `ReadWriteMany` access mode. However, I do like the concise way of stating this, which is similar to my original statement. I updated the statement because of a peer review comment that I received previously that it might be good to mention the default option so that the customer can distinguish the two, I guess. I'll return to the original statement that keeps things simplified.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
jself-sudoku: Well, you have just cleared up the thing I was most confused about. My technological gap had me unsure about it, but since you're are knowledgeable about RWX, it clears things up. There were too many "or" conjunctions, which was muddying the waters, so to speak. Updating accordingly.

downstream/modules/platform/con_storage-options-for-operator-installation-on-ocp.adoc
jself-sudoku: I just added OpenShift Data Foundation to the attributes file yesterday, per @ariordan-redhat 's request. So this is now updated.

downstream/modules/platform/proc_provision-ocp-storage-with-readwritemany-access-mode.adoc
```diff
@@ -0,0 +1,12 @@
...
+
+= Provisioning OCP storage with `ReadWriteMany` access mode
+
+To ensure successful installation of {OperatorPlatform}, set the storage type for {HubName} to `ReadWriteMany` access mode by completing the following steps.
+
+.Procedure
+
+. In your {HubName} instance, click the *File for Storage Type* drop-down list and select `ReadWriteMany`.
+
+. Click *Create*.
```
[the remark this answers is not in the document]
oraNod: ok, whatever you think but I do believe we make it clear that RWX is required for AAP.

ariordan-redhat (approved):

PR #349, opened 2022-06-07
AAP2033 - added quotes to secret values & other minor changes

dcdacosta:
Changes made for this PR include:

- Added quotes around secret values.
- Added note regarding unique database names in a single PostreSQL instance.
- Changed "Tower ingress type" to "Ingress type"

Preview link (VPN required): http://file.rdu.redhat.com/~ddacosta/aap2033/master.html

### DIFF
--- a/downstream/modules/platform/proc-operator-external-db-controller.adoc
+++ b/downstream/modules/platform/proc-operator-external-db-controller.adoc
@@ -6,7 +6,14 @@
 [role="_abstract"]
 For users who prefer to deploy {PlatformNameShort} with an external database, they can do so by configuring a secret with instance credentials and connection information, then applying it to their cluster using the `oc create` command.
 
-By default, the {PlatformName} operator automatically creates and configures a managed PostgreSQL pod in the same namespace as your {PlatformNameShort} deployment. A user may instead choose to use an external database if they prefer to use a dedicated node to ensure dedicated resources or to manually manage backups, upgrades, or performance tweaks. The following section outlines the steps to configure an external database for your {ControllerName} on a {PlatformNameShort} operator.
+By default, the {PlatformName} operator automatically creates and configures a managed PostgreSQL pod in the same namespace as your {PlatformNameShort} deployment. A user may instead choose to use an external database if they prefer to use a dedicated node to ensure dedicated resources or to manually manage backups, upgrades, or performance tweaks.
+
+[NOTE]
+====
+The same external database (PostgreSQL instance) can be used for both {HubName} and {ControllerName} as long as the database names are different. In other words, you can have multiple databases with different names inside a single PostgreSQL instance.
+====
+
+The following section outlines the steps to configure an external database for your {ControllerName} on a {PlatformNameShort} operator.
 
 .Prerequisite
 The external database must be a PostgreSQL database that is the version supported by the current release of {PlatformNameShort}.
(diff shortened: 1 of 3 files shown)


downstream/modules/platform/proc-operator-external-db-controller.adoc
```diff
@@ -6,7 +6,14 @@
 [role="_abstract"]
 For users who prefer to deploy {PlatformNameShort} with an external database, they can do so by configuring a secret with instance credentials and connection information, then applying it to their cluster using the `oc create` command.
 
-By default, the {PlatformName} operator automatically creates and configures a managed PostgreSQL pod in the same namespace as your {PlatformNameShort} deployment. A user may instead choose to use an external database if they prefer to use a dedicated node to ensure dedicated resources or to manually manage backups, upgrades, or performance tweaks. The following section outlines the steps to configure an external database for your {ControllerName} on a {PlatformNameShort} operator.
+By default, the {PlatformName} operator automatically creates and configures a managed PostgreSQL pod in the same namespace as your {PlatformNameShort} deployment. A user may instead choose to use an external database if they prefer to use a dedicated node to ensure dedicated resources or to manually manage backups, upgrades, or performance tweaks.
```
oraNod, line 9: The second sentence "A user may instead" sounds really confusing to me. I would advocate for addressing the reader directly so change "A user" to "You".

Also what is meant by "a dedicated node to ensure dedicated resources"? It seems like these two sentences could be collapsed a little. You could probably even condense this with the first sentence. That seems like it is slightly redundant with the "A user may" sentence, which addresses motivation.

Perhaps consider something like the following:

You can deploy {PlatformNameShort} with an external database instead of the managed PostgreSQL pod that the {PlatformName} operator automatically creates.
Using an external database lets you share and reuse resources and manually manage backups, upgrades, and performance optimizations.

dcdacosta: Thanks for the feedback @oraNod, I have implemented your suggested changes.

downstream/modules/platform/proc-operator-external-db-hub.adoc
```diff
@@ -6,7 +6,14 @@
 [role="_abstract"]
 For users who prefer to deploy {PlatformNameShort} with an external database, they can do so by configuring a secret with instance credentials and connection information, then applying it to their cluster using the `oc create` command.
 
-By default, the {PlatformName} operator automatically creates and configures a managed PostgreSQL pod in the same namespace as your {PlatformNameShort} deployment. A user may instead choose to use an external database if they prefer to use a dedicated node to ensure dedicated resources or to manually manage backups, upgrades, or performance tweaks. The following section outlines the steps to configure an external database for your {HubName} on a {PlatformNameShort} operator.
+By default, the {PlatformName} operator automatically creates and configures a managed PostgreSQL pod in the same namespace as your {PlatformNameShort} deployment. A user may instead choose to use an external database if they prefer to use a dedicated node to ensure dedicated resources or to manually manage backups, upgrades, or performance tweaks.
```
ariordan-redhat, line 9:
```suggestion
By default, the {PlatformName} operator automatically creates and configures a managed PostgreSQL pod in the same namespace as your {PlatformNameShort} deployment. You may instead choose to use an external database if you prefer to use a dedicated node to ensure dedicated resources or to manually manage backups, upgrades, or performance tweaks.
```

ariordan-redhat (commented):
Just one suggestion.
LGTM otherwise.

downstream/modules/platform/proc-operator-external-db-hub.adoc
ianf77: You might instead choose..

PR #351, opened 2022-06-09
[WIP] AAP-585 Importing CA certificates for automation mesh

TarikFD:
Jira: https://issues.redhat.com/browse/AAP-585

Documenting the procedure to import CA files to the inventory file during an installation to support a user's automation mesh certification. Adding a mention to these inventory file fields to the installation guide, but also adding more information on the automation mesh guide for the requirements and full instructions.

Preview (VPN required): http://file.rdu.redhat.com/kevchin/import-mesh-ca4/tmp/en-US/html-single/#proc-import-mesh-ca_installing-automation-mesh

### DIFF
--- a/downstream/assemblies/platform/assembly-installing-automation-mesh.adoc
+++ b/downstream/assemblies/platform/assembly-installing-automation-mesh.adoc
@@ -3,20 +3,24 @@ ifdef::context[:parent-context: {context}]
 
 
 [id="installing-automation-mesh"]
-= Installing {AutomationMesh}
+= Installing and configuring {AutomationMesh}
 
 
 :context: installing-automation-mesh
 
 
 [role="_abstract"]
-You can install an {AutomationMesh} using the instructions found in the {PlatformName} Installation Guide. {PlatformNameShort} installations are separated into unique scenarios based on the components deployed. You can incorporate {AutomationMesh} designs based on which platform components you intend to deploy in your environment. Review the systems requirements for each component and {AutomationMesh} node type before proceeding with your installation.
+You can install {AutomationMesh} using the instructions found in the link:https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.1/html/red_hat_ansible_automation_platform_installation_guide/index[{PlatformName} Installation Guide]. {PlatformNameShort} installations are separated into unique scenarios based on the components deployed. You can incorporate {AutomationMesh} designs based on which platform components you intend to deploy in your environment. Review the link:https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.1/html/red_hat_ansible_automation_platform_installation_guide/planning-installation#red_hat_ansible_automation_platform_system_requirements[{PlatformName} System Requirements] for each component and {AutomationMesh} node type before proceeding with your installation.
+
+The following sections provide options for you to configure your {AutomationMesh}, including an option to import your own CA certificate.
+
+include::platform/proc-import-mesh-ca.adoc[leveloffset=+1]
 
 
 [role="_additional-resources"]
-== Additional resources
+.Additional resources
 
-* See the {PlatformName} Reference Architecture for additional information before installing.
+* Review the link:https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.1/html/red_hat_ansible_automation_platform_installation_guide/planning-installation#red_hat_ansible_automation_platform_system_requirements[{PlatformName} System Requirements] for additional information before installing.
 
 ifdef::parent-context[:context: {parent-context}]
 ifndef::parent-context[:!context:]
--- a/downstream/assemblies/platform/assembly-installing-automation-mesh.adoc
+++ b/downstream/assemblies/platform/assembly-installing-automation-mesh.adoc
@@ -10,7 +10,7 @@ ifdef::context[:parent-context: {context}]
 
 
 [role="_abstract"]
-You can install {AutomationMesh} using the instructions found in the link:https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.1/html/red_hat_ansible_automation_platform_installation_guide/index[{PlatformName} Installation Guide]. {PlatformNameShort} installations are separated into unique scenarios based on the components deployed. You can incorporate {AutomationMesh} designs based on which platform components you intend to deploy in your environment. Review the link:https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.1/html/red_hat_ansible_automation_platform_installation_guide/planning-installation#red_hat_ansible_automation_platform_system_requirements[{PlatformName} System Requirements] for each component and {AutomationMesh} node type before proceeding with your installation.
+You can install {AutomationMesh} using the instructions found in the link:https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.1/html/red_hat_ansible_automation_platform_installation_guide/index[{PlatformName} Installation Guide]. {PlatformNameShort} installations are separated into unique scenarios based on the components deployed. You can incorporate {AutomationMesh} designs based on which platform components you intend to deploy in your environment.
 
 The following sections provide options for you to configure your {AutomationMesh}, including an option to import your own CA certificate.
 
--- a/downstream/modules/platform/proc-import-mesh-ca.adoc
+++ b/downstream/modules/platform/proc-import-mesh-ca.adoc
@@ -12,9 +12,14 @@ To import your own CA certificate, it must meet the following requirements:
 * asdasd
 
 .Procedure
-* In the `inventory` file, add the following variables and specify a path to both the certificate file (`.crt`) and the private RSA key (`.key`), following the example below:
+
+. Open the `inventory` file for editing.
+. Add the `mesh_ca_keyfile` variable and specify the full path to the CA certificate file (.crt).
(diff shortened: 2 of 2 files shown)


downstream/modules/platform/proc-import-mesh-ca.adoc
```diff
@@ -0,0 +1,29 @@
...
+NOTE: Users are not required to have a CA prior to their {AutomationMesh} installation. Users without a CA can proceed with their {PlatformNameShort} installation where the installer will generate a CA for the user using OpenSSL.
+
+.Prerequisites
+To import your own CA certificate, it must meet the following requirements:
+
+* asdasd
+* asdasd
+
+.Procedure
+* In the `inventory` file, add the following variables and specify a path to both the certificate file (`.crt`) and the private RSA key (`.key`), following the example below:
```
oraNod, line 15: I feel like this could be broken into steps.

. Open the `inventory` file for editing.
. Add the `mesh_ca_keyfile` variable and specify the full path to the CA certificate file (`.crt`).
. Add the `mesh_ca_certfile` variable and specify the full path to the private RSA key (`.key`).
. Save the changes to the `inventory` file.

TarikFD: This is a much elegant way to structure this. Thanks Don!

fosterseth:
I don't think we have specific requirements for the CA. I would only add a note explaining that the provided CA is going to be copied into each control node on the mesh and stored on disk.

`mesh_node_certfile` this certificate does have requirements. Currently we use the same cert for the backend TCP connection as well as the virtual network quic connections. The quic connection certs have to be tailor made for receptor to utilize, as described here https://receptor.readthedocs.io/en/latest/tls.html#above-the-mesh-tls

@shanemcd currently the installer allows users to specify `mesh_node_certfile`, which will bypass having receptor generate the cert at installation time. But these certs are not going to have the right meta data (othername, oid, etc) that receptor needs.

We would either need to describe those extra steps to get the correct certs, or maybe just not document that option so that users don't do it.

Long term, I think we should generate separate certificates for the virtual mesh connections (where receptor names are validated), and the backend connections (where receptor names are NOT validated). Users could provide their own certs for the backend TCP connections, and we can generate internal use only certs for the virtual mesh connections.

shanemcd:
I agree that we shouldn't document `mesh_node_certfile` given the requirements/limitations @fosterseth pointed out. We should only show how to BYO CA.

downstream/modules/platform/proc-import-mesh-ca.adoc
```diff
@@ -0,0 +1,31 @@
+[id="proc-import-mesh-ca_{context}"]
+
+= Importing a Certificate Authority (CA) certificate
+
+A Certificate Authority (CA) is used to certify and sign execution nodes in an {AutomationMesh} environment. Users who wish to provide their own CA can do so by specifying the path to the certificate and the private RSA key file in the `inventory` file of their {PlatformName} installer.
```
fosterseth, line 5: `is used to certify and sign execution nodes` might not be the right terminology. 

maybe `is used to verify and sign individual node certificates in an {AutomationMesh} environment`

downstream/modules/platform/proc-import-mesh-ca.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+.Example
+[subs="+quotes"]
+----
+[all:vars]
+mesh_ca_keyfile=/tmp/__<mesh_CA>__.key
+mesh_ca_certfile=/tmp/__<mesh_CA>__.crt
+----
+
+With the CA files added to the inventory file, run the installer to apply the CA. This process copies and stores the CA on each control node on your network. By default, the certfile and keyfile will be added to `/etc/receptor/tls/ca/`.
```
fosterseth, line 24: My mistake, it will copy the CA to each node (control and execution nodes) on the mesh.

downstream/modules/platform/proc-import-mesh-ca.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+.Optional: Assigning CA certificates to individual nodes
```
fosterseth, line 26: the usage of `mesh_node_certfile` should not be documented at this time.

downstream/modules/platform/proc-import-mesh-ca.adoc
```diff
@@ -0,0 +1,24 @@
+[id="proc-import-mesh-ca_{context}"]
+
+= Importing a Certificate Authority (CA) certificate
+
+A Certificate Authority (CA) is used to verify and sign individual node certificates in an {AutomationMesh} environment. Users who wish to provide their own CA can do so by specifying the path to the certificate and the private RSA key file in the `inventory` file of their {PlatformName} installer.
+
+NOTE: Users are not required to have a CA prior to their {AutomationMesh} installation. Users without a CA can proceed with their {PlatformNameShort} installation where the installer will generate a CA for the user using OpenSSL.
```
fosterseth, line 7: again, my mistake, but we don't use openssl as a backend. It uses the golang crypto library to generate CA, certs, and keys. I think we can just end the statement with "for the user".

fosterseth (approved):

downstream/modules/platform/proc-import-mesh-ca.adoc
```diff
@@ -0,0 +1,24 @@
...
+= Importing a Certificate Authority (CA) certificate
+
+A Certificate Authority (CA) is used to verify and sign individual node certificates in an {AutomationMesh} environment. Users who wish to provide their own CA can do so by specifying the path to the certificate and the private RSA key file in the `inventory` file of their {PlatformName} installer.
+
+NOTE: Users are not required to have a CA prior to their {AutomationMesh} installation. Users without a CA can proceed with their {PlatformNameShort} installation where the installer will generate a CA for the user.
+
+.Procedure
+
+. Open the `inventory` file for editing.
+. Add the `mesh_ca_keyfile` variable and specify the full path to the CA certificate file (.crt).
```
oraNod, line 12:
```suggestion
. Add the `mesh_ca_keyfile` variable and specify the full path to the CA certificate file (`.crt`).
```

downstream/modules/platform/proc-import-mesh-ca.adoc
```diff
@@ -0,0 +1,24 @@
...
+. Add the `mesh_ca_certfile` variable and specify the full path to the private RSA key (.key).
```
oraNod, line 13:
```suggestion
. Add the `mesh_ca_certfile` variable and specify the full path to the private RSA key file (`.key`).
```

downstream/modules/platform/proc-import-mesh-ca.adoc
oraNod: Could this also be a `cacerts` file?

downstream/modules/platform/proc-import-mesh-ca.adoc
```diff
@@ -0,0 +1,24 @@
+[id="proc-import-mesh-ca_{context}"]
+
+= Importing a Certificate Authority (CA) certificate
+
+A Certificate Authority (CA) is used to verify and sign individual node certificates in an {AutomationMesh} environment. Users who wish to provide their own CA can do so by specifying the path to the certificate and the private RSA key file in the `inventory` file of their {PlatformName} installer.
```
oraNod, line 5:
```suggestion
A Certificate Authority (CA) is used to verify and sign individual node certificates in an {AutomationMesh} environment. You can provide your own CA by specifying the path to the certificate and the private RSA key file in the `inventory` file of your {PlatformName} installer.
```

downstream/assemblies/platform/assembly-installing-automation-mesh.adoc
```diff
@@ -3,20 +3,24 @@ ifdef::context[:parent-context: {context}]
...
 
 
 [role="_abstract"]
-You can install an {AutomationMesh} using the instructions found in the {PlatformName} Installation Guide. {PlatformNameShort} installations are separated into unique scenarios based on the components deployed. You can incorporate {AutomationMesh} designs based on which platform components you intend to deploy in your environment. Review the systems requirements for each component and {AutomationMesh} node type before proceeding with your installation.
+You can install {AutomationMesh} using the instructions found in the link:https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.1/html/red_hat_ansible_automation_platform_installation_guide/index[{PlatformName} Installation Guide]. {PlatformNameShort} installations are separated into unique scenarios based on the components deployed. You can incorporate {AutomationMesh} designs based on which platform components you intend to deploy in your environment.
+
+The following sections provide options for you to configure your {AutomationMesh}, including an option to import your own CA certificate.
+
+include::platform/proc-import-mesh-ca.adoc[leveloffset=+1]
```
oraNod, line 17: To be honest this assembly seems a little odd. It looks like it was created as a sort of "stub" assembly that didn't contain any include directives.

As a user the title doesn't quite match the abstract, which is basically a pointer to the installation guide. There's no procedure for installation in the user story so the high-level heading doesn't line up with the content.

It seems to me like we're kind of "bolting on" more content here and this assembly will turn into a hodge-podge sort of dump. Also, when I look at the title for the mesh docs it appears that this assembly is at the end, after deprovisioning.

What about moving this under the planning assembly before the example topologies? You could rename the assembly title to something like "Setting Up {AutomationMesh}". It would then make sense to me to break the first couple sentences into a new concept topic and expand out a little about installing mesh. Then the procedure about adding the CA cert.

TarikFD:
Redoing this PR with #383 to resolve local build issues.

PR #354, opened 2022-06-10
AAP1808 - Added information for AWS storage class prereq

dcdacosta:
Fixes issue #https://issues.redhat.com/browse/AAP-1808

Changes proposed in this pull request:

- Added AWS Storage Class requirements for RWX.

Link to preview (requires VPN):  http://file.rdu.redhat.com/~ddacosta/aapoperator_1808/master.html

### DIFF
--- a/downstream/assemblies/platform/assembly-install-aap-operator.adoc
+++ b/downstream/assemblies/platform/assembly-install-aap-operator.adoc
@@ -11,6 +11,10 @@ ifdef::context[:parent-context: {context}]
 
 .Prerequisites
 * You have installed the {PlatformName} catalog in Operator Hub.
+* You have created a Storage Class if one does not exist. See link:https://docs.openshift.com/container-platform/4.10/storage/dynamic-provisioning.html[Dyamic Provisioning] for details.
+* To run {OCP} clusters on Amazon Web Services with ReadWriteMany, you must add NFS or other storage. See the following for details:
+** For information on the AWS Elastic Block Store (EBS) or to use the 'aws-ebs' storage class, see link:https://docs.aws.amazon.com/eks/latest/userguide/storage-classes.html[Storage Classes]
+** To use multi-attach ReadWriteMany for AWS EBS, see link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volumes-multi.html[Attaching a volume to multiple instances with Amazon EBS Multi-Attach]
 
 include::platform/proc-install-aap-operator.adoc[leveloffset=+2]
 
--- a/downstream/assemblies/platform/assembly-install-aap-operator.adoc
+++ b/downstream/assemblies/platform/assembly-install-aap-operator.adoc
@@ -11,10 +11,10 @@ ifdef::context[:parent-context: {context}]
 
 .Prerequisites
 * You have installed the {PlatformName} catalog in Operator Hub.
-* You have created a Storage Class if one does not exist. See link:https://docs.openshift.com/container-platform/4.10/storage/dynamic-provisioning.html[Dyamic Provisioning] for details.
-* To run {OCP} clusters on Amazon Web Services with ReadWriteMany, you must add NFS or other storage. 
-** For information on the AWS Elastic Block Store (EBS) or to use the `aws-ebs` storage class, see link:https://docs.aws.amazon.com/eks/latest/userguide/storage-classes.html[Storage Classes].
-** To use multi-attach ReadWriteMany for AWS EBS, see link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volumes-multi.html[Attaching a volume to multiple instances with Amazon EBS Multi-Attach].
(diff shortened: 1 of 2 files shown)


downstream/assemblies/platform/assembly-install-aap-operator.adoc
```diff
@@ -11,6 +11,10 @@ ifdef::context[:parent-context: {context}]
 
 .Prerequisites
 * You have installed the {PlatformName} catalog in Operator Hub.
+* You have created a Storage Class if one does not exist. See link:https://docs.openshift.com/container-platform/4.10/storage/dynamic-provisioning.html[Dyamic Provisioning] for details.
+* To run {OCP} clusters on Amazon Web Services with ReadWriteMany, you must add NFS or other storage.
```
oraNod, line 15: Where does NFS come from? This reads a little strange to me because RWX access mode is specific to how pods on the OCP cluster interacts with the persistent volume. The existence of RWX access mode implies there is a persistent volume. So it's like the wording is the other way round.

Also this sounds like a pre-req for OCP itself rather than AAP. Maybe it could be something more like:

You have added a persistent volume with the `ReadWriteMany` access mode.

downstream/assemblies/platform/assembly-install-aap-operator.adoc
```diff
@@ -11,6 +11,10 @@ ifdef::context[:parent-context: {context}]
 
 .Prerequisites
 * You have installed the {PlatformName} catalog in Operator Hub.
+* You have created a Storage Class if one does not exist. See link:https://docs.openshift.com/container-platform/4.10/storage/dynamic-provisioning.html[Dyamic Provisioning] for details.
+* To run {OCP} clusters on Amazon Web Services with ReadWriteMany, you must add NFS or other storage. 
+** For information on the AWS Elastic Block Store (EBS) or to use the `aws-ebs` storage class, see link:https://docs.aws.amazon.com/eks/latest/userguide/storage-classes.html[Storage Classes].
+** To use multi-attach ReadWriteMany for AWS EBS, see link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volumes-multi.html[Attaching a volume to multiple instances with Amazon EBS Multi-Attach].
```
oraNod, line 17: Do we really want to point to 3rd party docs? I don't think that is recommended.

OCP docs have lots of platform specific details, including AWS: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html-single/storage/index#persistent-storage-aws

downstream/assemblies/platform/assembly-install-aap-operator.adoc
```diff
@@ -11,6 +11,10 @@ ifdef::context[:parent-context: {context}]
 
 .Prerequisites
 * You have installed the {PlatformName} catalog in Operator Hub.
+* You have created a Storage Class if one does not exist. See link:https://docs.openshift.com/container-platform/4.10/storage/dynamic-provisioning.html[Dyamic Provisioning] for details.
```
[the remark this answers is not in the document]
ariordan-redhat: Do we need to specify which StorageClass object definition to use?

ariordan-redhat (commented):
Left a comment that's more of a question.

downstream/assemblies/platform/assembly-install-aap-operator.adoc
dcdacosta: @oraNod The NFS information was previously part of a note placed just before the procedures. In the issue, it was stated that this is actually a pre-req and requested that all pre-reqs be included. 

These prerequisites are in the section for OCP so I think it's ok for it to come across that way, but I will update with your suggested change.

downstream/assemblies/platform/assembly-install-aap-operator.adoc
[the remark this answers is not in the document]
dcdacosta: @ariordan-redhat All good questions. According to Christian, he's not sure what the defaults SC should be and said we should point the users to these sources in the short term and then "Long term, it would be good for us to take the time to figure out a recommended path and document steps."

downstream/assemblies/platform/assembly-install-aap-operator.adoc
oraNod: Thanks @dcdacosta I hope I'm not adding noise with my review comments. I just think like it could be clarified a little bit more. I find looking at the procedures in the OCP docs helped me understand a little more: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html-single/storage/index#storage-create-storage-class_persistent-storage-aws 

The first procedure is about creating a `StorageClass` then creating a PVC. So maybe we could make it a single pre-req like:

You have created the appropriate `StorageClass` object for your platform and a persistent volume claim (PVC) with `ReadWriteMany` access mode.

Anyway I'll leave you to it but do hope this is helpful in some way. Cheers.

rooftopcellist (approved):

PR #363, opened 2022-06-15
add aap-common content

oraNod:
Creates an AAP common content branch that includes a shared assembly and inclusive language module that we can sync to downstream branches.


ariordan-redhat (approved):
LGTM

PR #367, opened 2022-06-16
resolves #366 inclusive language statement

### LINKED ISSUE #366: Add inclusive language statement to titles
Update the titles to add the inclusive language statement from aap-common.

oraNod:
Fixes #366 by adding the inclusive language statement from aap-common to downstream titles.


oraNod:
@dcdacosta Now that we've got the downstream sync jobs running and `aap-common` is set up, I'm going to add the inclusive language statement to titles.

I didn't add the module to each and every title though. In some of the smaller sub-directories it kind of gets in the way so I thought we might just include it for the "main" books, if that makes any sense.

Note that the include directive is commented because `aap-common` is synced downstream and it would throw a build error in this repo. I'll set things up so the module gets included downstream. I don't think this is ideal but until we separate upstream and downstream a little more it's the easiest way.

Thanks.

ariordan-redhat (approved):
LGTM

dcdacosta (approved):
/lgtm

Thanks Don!

PR #371, opened 2022-06-23
create asciidoc alts for master.adoc

oraNod:
This PR removes `master.adoc` and `docinfo.xml` artifacts for the AAP on Clouds titles. These artifacts now exist in the downstream GitLab repo only because they are required for the publishing system. 

We should keep title artifacts downstream only so it is easier for Felicia to make changes to resolve publishing issues, as we observed recently with the new splash page. Additionally the `master.adoc` file name uses problematic language but is required by the publishing system. So it would be better, imo, not to expose this in the upstream repo.

To replace the `master.adoc` file we can use title files with the `.asciidoc` extension instead. This means we can give titles in the upstream more meaningful names. Additionally, using the `.asciidoc` extension distinguishes the title files from all the other `.adoc` files in the repository, which gives us the ability to iterate over the repo and generate HTML from all the title files.

Another change in this PR is to use `stories.adoc` to include the assemblies / user stories that each title contains. Having a separate file that references assemblies provides more logical organization and granularity. Additionally it means we can sync the `stories.adoc` files to GitLab if we want to use the same set of assemblies for the downstream title.

Finally I've added some macros that we can use in the upstream, like `sectanchors`, that are convenient but not supported downstream.

I've also updated the instructions for building titles in the README.


ariordan-redhat (approved):
I built the Azure doc with asciidoctor and it looks OK.
The only issue I can see with this change is that the build that you get with bccutil (which requires master.adoc and docinfo.xml) has an index for the doc and uses Red Hat style.

oraNod:
> I built the Azure doc with asciidoctor and it looks OK. The only issue I can see with this change is that the build that you get with bccutil (which requires master.adoc and docinfo.xml) has an index for the doc and uses Red Hat style.

Right. I didn't think anyone was using bccutil to build titles in this repo. It seems like that's something we should be doing in the GitLab repo with the downstream only titles.

From the review docs that I've seen on the fileserver it looks like everyone was building with asciidoctor rather than bccutil.

oraNod:
@ariordan-redhat I'm going to close this PR for now in the interest of avoiding too many changes. Also there are now a bunch of conflicts and this PR seems to be messed up.

One thing to keep in mind for the Ansible on Clouds docs is that the `master.adoc` file is not used in the downstream repo. So if you add an assembly or change the order then we need to make sure those changes are reflected downstream.

PR #381, opened 2022-06-28
Added Python requirement to AAP docs

ianf77:
Added requirement for Python 3.8 or later to Installation guide.
The Upgrade/Miration Guide has no correspponding requirements section so
that document has been missed out.

Add python requirements for AAP into docs

https://issues.redhat.com/browse/AAP-3385

### DIFF
--- a/downstream/modules/platform/ref-system-requirements.adoc
+++ b/downstream/modules/platform/ref-system-requirements.adoc
@@ -23,6 +23,8 @@ h| Ansible | version 2.9 required |
 
 h| RAM | 4Gb minimum |
 
+h| Python | 3.8 or later |
+
 * 4 GB RAM (minimum and recommended for Vagrant trial installations)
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-system-requirements.adoc
```diff
@@ -23,6 +23,8 @@ h| Ansible | version 2.9 required |
 
 h| RAM | 4Gb minimum |
 
+h| Python | 3.8 or later |
+
```
ariordan-redhat, line 27:
```suggestion

h| Python | 3.8 or later |

h| RAM | 4Gb minimum |

```

ariordan-redhat: @ianf77 I swapped the RAM and Python lines to make the table render properly in the build.

PR #409, opened 2022-07-01
Added disk space requirements

ianf77:
Added disk space requirements for disk service node.
Also added requirement for 20 GB to be allocated to /home/awx.
As it was a note I've removed the note for Automation controller and made it
a heading. Less messy that way.

Disk space requirements for EE's not documented

https://issues.redhat.com/browse/AAP-4123

### DIFF
--- a/downstream/modules/platform/ref-system-requirements.adoc
+++ b/downstream/modules/platform/ref-system-requirements.adoc
@@ -30,12 +30,16 @@ h| Python | 3.8 or later |
 
 
 
-.Automation controller
+== Automation controller
+
+{ControllerNameStart} is a distributed system, where different software components can be colocated or distributed across multiple compute nodes.
+In the installer, node types of control, hybrid, execution, and hop are provided as abstractions to help the user design the topology appropriate for their use case.
+Recommendations for node sizing are provided in the following table.
 
 [NOTE]
-=====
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-system-requirements.adoc
```diff
@@ -30,12 +30,16 @@ h| Python | 3.8 or later |
 
 
 
-.Automation controller
+== Automation controller
+
+{ControllerNameStart} is a distributed system, where different software components can be colocated or distributed across multiple compute nodes.
```
oraNod, line 35:
```suggestion
{ControllerNameStart} is a distributed system, where different software components can be co-located or deployed across multiple compute nodes.
```

downstream/modules/platform/ref-system-requirements.adoc
```diff
@@ -30,12 +30,16 @@ h| Python | 3.8 or later |
...
+In the installer, node types of control, hybrid, execution, and hop are provided as abstractions to help the user design the topology appropriate for their use case.
+Recommendations for node sizing are provided in the following table.
```
oraNod, line 37:
```suggestion
The following table provides recommendations for node sizing:
```

downstream/modules/platform/ref-system-requirements.adoc
oraNod: avoid duplication of "distributed" - seems like you're using the same word to describe itself. 

also not 100% but I feel like there should be a hyphen in `co-located` otherwise it looks like a misspelling of "collocated"

PR #415, opened 2022-07-06
AAP-4359 Maintain Hub API during AAP upgrade + basic upgrade section

TarikFD:
Jira: https://issues.redhat.com/browse/AAP-4359

Request was to document feature in the installer for users to import an existing Hub API token or generate a new one using the AAP installer during an AAP upgrade. In doing so, I'm adding a section to outline a "basic upgrade" of AAP to guide users through editing the inventory file and running the setup script. Working with product team to ensure this new section makes sense.

Preview (VPN required): http://file.rdu.redhat.com/kevchin/aap-upgrade6/tmp/en-US/html-single/

PR #421, opened 2022-07-12
Updated title and inventory file

ianf77:
Changed title of Section 3.1.4 to reflect the actual content.
Removed AAP from the tile so as not to be too long.
Edited inventory file as per instructions in the ticket.

Typo in the title of 3.1.4 for AAP2  documentation

https://issues.redhat.com/browse/AAP-4678

### DIFF
--- a/downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
+++ b/downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
@@ -2,11 +2,12 @@
 
 [id="ref-standlone-platform-inventory_{context}"]
 
-= Example {PlatformName} inventory file with an external managed database
+= Example inventory file for a database on the automation controller node or a non-installer managed database
 
 
 [role="_abstract"]
-This example describes how you can populate the inventory file to install {PlatformName}. This installation inventory file includes both {ControllerName} and {HubName} with a database on the {ControllerName} node or non-installer managed database.
+This example describes how you can populate the inventory file to install {PlatformName}. 
+This installation inventory file includes both {ControllerName} and {HubName} with a database on the {ControllerName} node or non-installer managed database.
 
 [IMPORTANT]
 ====
@@ -24,8 +25,8 @@ automationhub.acme.org
 
 [all:vars]
 admin_password='<password>'
-pg_host='database-01.acme.org'
-pg_port='5432'
+pg_host=''
+pg_port=''
 pg_database='awx'
 pg_username='awx'
 pg_password='<password>'
@@ -38,7 +39,7 @@ registry_password='<registry password>'
 # Automation Hub Configuration
 #
 automationhub_admin_password='<password>'
-automationhub_pg_host='database-01.acme.org'
+automationhub_pg_host='controller'
 automationhub_pg_port='5432'
 automationhub_pg_database='automationhub'
 automationhub_pg_username='automationhub'
--- a/downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
+++ b/downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
@@ -6,7 +6,7 @@
 
 
 [role="_abstract"]
-This example describes how you can populate the inventory file to install {PlatformName}. 
+This example describes how you can populate the inventory file to install {PlatformName}.
 This installation inventory file includes both {ControllerName} and {HubName} with a database on the {ControllerName} node or non-installer managed database.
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
```diff
@@ -2,11 +2,12 @@
 
 [id="ref-standlone-platform-inventory_{context}"]
 
-= Example {PlatformName} inventory file with an external managed database
+= Example inventory file for a database on the automation controller node or a non-installer managed database
```
oraNod, line 5: Super long title that sounds like it covers most scenarios. How about shortening to something like "Database inventory file example"?

downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
```diff
@@ -2,11 +2,12 @@
...
 
 
 [role="_abstract"]
-This example describes how you can populate the inventory file to install {PlatformName}. This installation inventory file includes both {ControllerName} and {HubName} with a database on the {ControllerName} node or non-installer managed database.
+This example describes how you can populate the inventory file to install {PlatformName}.
```
oraNod, line 9:
```suggestion
This example illustrates how to build an inventory file when installing {PlatformName}. 
```

downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
```diff
@@ -38,7 +39,7 @@ registry_password='<registry password>'
 # Automation Hub Configuration
 #
 automationhub_admin_password='<password>'
-automationhub_pg_host='database-01.acme.org'
+automationhub_pg_host='controller'
```
oraNod, line 42:
```suggestion
automationhub_pg_host='<controller>'
```

oraNod: Replaceable value.

downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
```diff
@@ -38,7 +39,7 @@ registry_password='<registry password>'
...
 automationhub_pg_port='5432'
 automationhub_pg_database='automationhub'
 automationhub_pg_username='automationhub'
```
oraNod, line 45: Do you think you could put the example in a `ini/database.ini` file then reference it with an includes directive? 

Similar to what Kevin is doing here: https://github.com/RedHatInsights/red-hat-ansible-automation-platform-documentation/pull/415/files#diff-64084273a8eab3e0c8b734bd9131033536167c8a0ad2cae15ff72993ee6a08dd

downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
adonisgarciac: It is not a database inventory file example. It is a inventory file to install AAP. I think that the proposed title is correct  because it follow the structure of the other sections.

downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
adonisgarciac: I don't understand why this has to be changed. This sentence, in my opinion, is totally correct right now: 
This example describes how you can populate the inventory file to install {PlatformName}. This installation inventory file includes both {ControllerName} and {HubName} with a database on the {ControllerName} node or non-installer managed database.

downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
adonisgarciac, line 42:
```suggestion
automationhub_pg_host='tower.acme.org'
```
I think that tower.acme.org is more suitable because as you can see, this is the automationcontroller host in this inventory example:  https://github.com/RedHatInsights/red-hat-ansible-automation-platform-documentation/blame/main/downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc#L19-L20

downstream/modules/platform/ref-platform-non-inst-database-inventory.adoc
oraNod: Thanks for keeping me honest @adonisgarciac I'll delete my suggestion if it doesn't fit.

adonisgarciac (approved):
LGFM

PR #427, opened 2022-07-15
Removed instructions for installing builder with pip

ianf77:
Removed instructions from the builder guide about installing using pip

Remove Instructions on Installing ansible-builder via pip

https://issues.redhat.com/browse/AAP-4086


oraNod (approved):
LGTM and build is fine

PR #435, opened 2022-07-19
Update latest OpenShift version to 4.10

ariordan-redhat:
Update attribute for latest OpenShift version to 4.10

Affects attributes file.


oraNod (approved):
LGTM

PR #439, opened 2022-08-02
AAP-4999 Add attribute for RHSSO

ariordan-redhat:
Add an attribute for RHSSO to future-proof the AAP docs in case of product name changes.
AAP-4999

### DIFF
--- a/downstream/attributes/attributes.adoc
+++ b/downstream/attributes/attributes.adoc
@@ -101,3 +101,6 @@
 :OCPLatest: 4.10
 :ODF: Red Hat OpenShift Data Foundation
 :ODFShort: OpenShift Data Foundation
+
+// Red Hat products
+:RHSSO: Red Hat Single Sign-On
--- a/downstream/attributes/attributes.adoc
+++ b/downstream/attributes/attributes.adoc
@@ -104,3 +104,4 @@
 
 // Red Hat products
 :RHSSO: Red Hat Single Sign-On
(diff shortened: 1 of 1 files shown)


ariordan-redhat:
Should we have a separate attribute for the RHSSO Operator?  Ideally, we shouldn't need it.

This is just a suggestion, in case a name change for the RHSSO product isn't immediately reflected in the RHSSO Operator name on the OCP UI.

downstream/attributes/attributes.adoc
```diff
@@ -101,3 +101,7 @@
 :OCPLatest: 4.10
 :ODF: Red Hat OpenShift Data Foundation
 :ODFShort: OpenShift Data Foundation
+
+// Red Hat products
+:RHSSO: Red Hat Single Sign-On
+:RHSSOOperator: Red Hat Single Sign-On Operator
```
oraNod, line 107:
```suggestion
:OperatorRHSSO: Red Hat Single Sign-On Operator
```

oraNod: I think this should follow the convention of the other [Operator attributes](https://github.com/RedHatInsights/red-hat-ansible-automation-platform-documentation/blob/343e0a70fc5489d6f70faf5f04531985b2961346/downstream/attributes/attributes.adoc#L22)

ariordan-redhat: @oraNod Done. Thanks.

oraNod (approved):

PR #447, opened 2022-08-05
AAP-5037 Use attributes for SSO and SSO Operator

ianf77:
Replace references to RH Single Sign-On with attribute.

https://issues.redhat.com/browse/AAP-5037

### DIFF
--- a/downstream/modules/platform/proc-create-keycloak-realm.adoc
+++ b/downstream/modules/platform/proc-create-keycloak-realm.adoc
@@ -9,9 +9,9 @@ Realms are isolated from one another and can only manage and authenticate the us
 .Procedure
 
 . Navigate to menu:Operator[Installed Operators].
-. Select the *Red Hat Single Sign-On Operator* project.
+. Select the *{OperatorRHSSO}* project.
 . Select the *Keycloak Realm* tab and click btn:[Create Keycloak Realm].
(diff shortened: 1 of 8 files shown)


downstream/modules/platform/proc-create-keycloak-realm.adoc
```diff
@@ -44,7 +44,7 @@ spec:
 | metadata.labels.app: |Set labels to a unique value. This is used when creating the client CR.
 | metadata.labels.realm: | Set labels to a unique value. This is used when creating the client CR.
 | spec.realm.id: | Set the realm name and id. These must be the same.
-| spec.realm.realm | Set the realm name and id. These must be the same.
+| spec.realm.realm: | Set the realm name and id. These must be the same.
 | spec.realm.displayname: | Set the name to display.
```
oraNod, line 48: Nit that is unrelated to this PR but maybe worth changing. Shouldn't these fields be wrapped in single quotes?

e.g. `spec.realm.realm:`

ariordan-redhat (approved):
Substitutions all look good to me

downstream/modules/platform/proc-create-keycloak-realm.adoc
ariordan-redhat: @oraNod I created a Jira issue for the fields format: I'll create a PR for it.

Issue #448, opened 2022-08-05
Create bin folders for sync scripts


oraNod:
Externalize the bash sections of each Jenkins job to synchronize the content of this repository downstream.

oraNod:
There should be a separate and unique PR for each branch that gets sync'ed downstream.

PR #451, opened 2022-08-05
[2.2] Add bin folder with sync scripts

oraNod:
Fixes issue #448 for the 2.2 branch.

### DIFF
--- a/bin/sync_ocp_latest.sh
+++ b/bin/sync_ocp_latest.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# Set the path to the file that contains the ":OCPLatest:" attribute.
+target=target/downstream/attributes/attributes.adoc
+
+# Retrieve the OCP attributes file from the OpenShift docs repository.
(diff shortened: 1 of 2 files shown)


bin/sync_ocp_latest.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Set the path to the file that contains the ":OCPLatest:" attribute.
+target=target/downstream/attributes/attributes.adoc
```
ariordan-redhat, line 4: Just wondering if the code would be easier to read if you used `attributes=` instead of `target=`

ariordan-redhat (approved):
LGTM

bin/sync_ocp_latest.sh
oraNod: yeah, good point. I kept it as `target` to be consistent with the other scripts but maybe that's not really necessary. `attributes` would be a little more descriptive here. plus the fact there is no source variable like the other scripts.

PR #464, opened 2022-08-08
2.2 AAP-5037 Use attributes for SSO and SSO Operator (#447)

ianf77:
Replace references to RH Single Sign-On with attribute.

https://issues.redhat.com/browse/AAP-5037


ariordan-redhat (approved):
LGTM

PR #470, opened 2022-08-16
Update broken hub link

ianf77:
Updating this again to replace version with attribute as the stated version
is rather old. This will be backported to 2.1 and 2.2.

Broken Links found for version 1.0 and 1.2 for ANSIBLE GALAXY COLLECTIONS IN AUTOMATION HUB

https://issues.redhat.com/browse/AAP-1625


ianf77:
I wonder about the version of geerlinguy.php_roles. Is it still v 0.9.3?

ariordan-redhat:
> 

I think the `geerlingguy` role is just used as an example of a collection to sync, so grabbing the most recent version isn't necessary for this doc. It would probably be a good idea to change the example at some point, though.

ariordan-redhat (approved):
LGTM

PR #471, opened 2022-08-17
AAP-5260 Add virtual appliance routing info

ariordan-redhat:
AAP-5260
Add info about configuring routing through an Azure Virtual Appliance

so that

users can configure routes to allow traffic from their Ansible on Azure instance through an Azure Virtual Appliance to quay and registry.redhat.io, as well as to automation targets.

Affects titles/aap-on-azure


ianf77 (approved):
LGTM

ariordan-redhat:
#476  backports to 2.1
#477  backports to 2.2

PR #475, opened 2022-08-19
AAP-5260 Update virtual appliance firewall content

ariordan-redhat:
AAP-5260
Update info about configuring routing through an Azure Virtual Appliance firewall

so that

outbound traffic from the full CIDR range of the managed application can access quay and registries, and so that you can enable all outbound traffic from port 443

### DIFF
--- a/downstream/modules/aap-on-azure/proc-azure-update-route-tables.adoc
+++ b/downstream/modules/aap-on-azure/proc-azure-update-route-tables.adoc
@@ -47,7 +47,7 @@ After you have configured the routing rules, traffic is routed to and from {Plat
 
 .Outbound routing through virtual appliances
 
-If your organization uses Azure firewall services or third-party firewall appliances through a Virtual Appliance connection, you must configure outbound connectivity from the managed application, to allow Red Hat to maintain your application and to allow automation against external resources.
+If your organization uses Azure firewall services or third-party firewall appliances through a Virtual Appliance connection, you must configure outbound connectivity from the managed application, to enable Red Hat to maintain your application and to enable automation against external resources.
 
 The easiest way to implement this is to create a firewall rule that allows all outbound traffic from port 443.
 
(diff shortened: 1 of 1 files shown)


downstream/modules/aap-on-azure/proc-azure-update-route-tables.adoc
```diff
@@ -47,10 +47,14 @@ After you have configured the routing rules, traffic is routed to and from {Plat
 
 .Outbound routing through virtual appliances
 
-If your organization uses Azure firewall services or third-party firewall appliances through a Virtual Appliance connection, you must configure routes to enable application management and automation against external resources.
+If your organization uses Azure firewall services or third-party firewall appliances through a Virtual Appliance connection, you must configure outbound connectivity from the managed application, to allow Red Hat to maintain your application and to allow automation against external resources.
```
ianf77, line 50: IBM style guide says not to use 'allow', so maybe:
so that Red Hat can manage your application...
and
...a firewall rule that routes all outbound traffic from port 443.

ariordan-redhat: As this would be a firewall allow rule, do you think it would be OK in this instance?

ariordan-redhat: Because firewalls allow and block traffic, rather than route traffic, as far as I know.

ariordan-redhat: Changed "allow" to enable in line 50

ariordan-redhat: @ianf77 For the firewall, I think we should stick with "allow" as firewall rules block or allow traffic: they don't route it.

ariordan-redhat:
#476 backports to 2.1
#477 backports to 2.2

PR #478, opened 2022-08-19
AAP-5310 Add attribute for Ansible Galaxy

ariordan-redhat:
AAP-5310 
Add an attribute to represent Ansible Galaxy
Backport to 2.1 and 2.2


rogrange:
I did a fairly comprehensive search and I continuously received "Ansible Galaxy" as a result. Even the Red Hat Ansible Galaxy searches came up with "Ansible Galaxy.," So at this point I would be fairly confident in the current attribute.

ianf77 (approved):
LGTM

PR #479, opened 2022-08-19
AAP-5311 Update latest OCP version to 4.11

ariordan-redhat:
Update the attribute representing the latest OpenShift release to 4.11
AAP-5311


rogrange:
This appears to be correct.

PR #488, opened 2022-08-23
AAP-5335 Fix ID for tools assembly

ariordan-redhat:
AAP-5335

Affects Platform Creator Guide (`/titles/dev-guide/`)

Chapter 4 isn't rendering on the customer portal: clicking on the chapter name in the left nav causes a 404-not-found error.
You can navigate to all other chapters from the left nav.

The local build is OK.

in the Pantheon prod build, you can scroll to the chapter, but clicking the chapter name in the left nav doesn't navigate to the chapter title.
Clicking all other chapters in the left nav brings you to the chapter.


ianf77 (approved):
LGTM

PR #491, opened 2022-08-23
2.1 AAP-5335 Fix ID for tools assembly (#488)

ariordan-redhat:
Backports #488 to 2.1

AAP-5335

Affects Platform Creator Guide (`/titles/dev-guide/`)


ianf77 (approved):
LGTM

PR #498, opened 2022-08-24
AAP-5346 Added AKS node pool resource group info

ariordan-redhat:
AAP-5346
Affects the Ansible on Azure guide (`/titles/aap-on-azure/`)
Incorporate information from a [KB article](https://access.redhat.com/articles/6869301  in the product documentation.

So that
customers understand the node pool resource group that's created when Ansible on Azure is deployed.


lmalivert:
LGTM @ariordan-redhat

PR #535, opened 2022-09-13
AAP-5740 Add attribute for console.redhat.com

ianf77:
Add atttribute for console.redhat.com

https://issues.redhat.com/browse/AAP-5740


ariordan-redhat (approved):
LGTM

PR #547, opened 2022-09-26
AAP-2123 Improve automation mesh examples

ariordan-redhat:
AAP-2123

Add examples for Automation mesh and explain the context.


saito-hideki:
@ariordan-redhat I have checked this PR and have quick question about the topology in the control nodes(For example, downstream/images/mesh-multiple-hybrid-nodes.png).
If the multiple hybrid nodes are configured in the inventory like below:

```ini
[automationcontroller]
aap_c_1.example.com
aap_c_2.example.com
aap_c_3.example.com
```

It is my understanding that the network topology of the automation mesh will be a mesh configuration like below rather than a one-way aap_c_3->aap_c_1->aap_c_2 as in the diagram above.

controller_mesh


If my understanding is wrong, please ignore this comment :)

ariordan-redhat:
Images updated. Do you have any further comments on this PR, @saito-hideki ?

ianf77 (commented):
Additional comment, to me, in the mesh-multi-hop.png the blue curved line gets a bit lost, it is possible to separate it from the lines above

downstream/modules/platform/proc-defining-node-types.adoc
```diff
@@ -5,7 +5,7 @@
 = Defining {AutomationMesh} node types
 
 [role="_abstract"]
-You can a define node type either by its default value assigned by the inventory group or by using the `node_type` host variable. Specify the `node_type` either as part of the inventory group or within the inventory `vars` group.  This section provides examples of how you can define node types in your inventory file.  Nodes in `[execution_nodes]` default execution node_type. Hybrid node types can be overridden to be control type via `node_type=control`. Execution node type can be overridden to be hope node type via `node_type=hop`
+You can a define node type either by its default value assigned by the inventory group or by using the `node_type` host variable. Specify the `node_type` either as part of the inventory group or within the inventory `vars` group.  This section provides examples of how you can define node types in your inventory file.  Nodes in `[execution_nodes]` default execution node_type. Hybrid node types can be overridden to be control type via `node_type=control`. Execution node type can be overridden to be hop node type via `node_type=hop`
```
ianf77, line 8: Instead of via `note_type=control', say 'using the `node_type=control` variable.
Also type via 'node_type=hop', say 'type using `node_type=hop'

I've documented node_type as an inventory file variable and it would be node_type = 'hop'. Is that worth checking?

ariordan-redhat: @ianf77 I just corrected a typo in this paragraph. I think it needs restructuring.

ariordan-redhat:
> Additional comment, to me, in the mesh-multi-hop.png the blue curved line gets a bit lost, it is possible to separate it from the lines above

Unfortunately not, if you use the graph generated by Graphviz!  The arrows are generated automatically. I messed around with the vertical spacing, but that didn't help much. I also tried to move the blue arrow to the bottom of the control nodes, but then it did a big swoop around the execution nodes at the bottom of the diagram, which looked weird.

I have some more work to do on the diagrams: it might make sense to omit the blue arrow in the rendered graph, and draw it in with an image editor.

ariordan-redhat:
The user can generate these graphs by running a command that processes the inventory file, so I thought it would make sense to use images that show the output that that users would see if they ran the command themselves.

ianf77 (approved):
LGTM

PR #568, opened 2022-10-11
AAP-1505 Fix build error

ariordan-redhat:
Fix build error - Operator installation guide couldn't be built because one of the files included in an assembly was mis-spelt.


ianf77 (approved):
LGTM

PR #613, opened 2022-11-09
AAP-1649: Adopt new Pulp RBAC (#587)

lmalivert:
Jira: [AAP-1649](https://issues.redhat.com/browse/AAP-1649)

- Users can now add specific permissions to roles, and then add those roles to groups.

- Added instructions for assigning roles from both the Groups menu and the Namespaces menu.


ianf77 (approved):
LGTM

PR #626, opened 2022-11-17
2.2 Updated with two short notes for single and multi node cluster instal…

ariordan-redhat:
Backports #583 to 2.2:

* Updated with two short notes for single and multi node cluster installation.


ianf77 (approved):
LGTM

PR #648, opened 2022-11-25
2.3 Fix build issues for 2.3 docs (#647)

ariordan-redhat:
Title mismatches for 2 documents were causing Pantheon builds to fail.
The titles in `master.adoc` and `docinfo.xml` must both match the title in Pantheon. 
Titles are case-sensitive.


emurtoug (approved):
LGTM

PR #658, opened 2022-11-29
AAP-1254a Modify links in installation guide

ariordan-redhat:
Fix links in the inventory guide.
Issue arose because building with asciidoctor doesn't pick up some errors that break Pantheon builds.


rogrange (approved):
The xref changes look good, and the other changes are appropriate.

PR #677, opened 2022-12-06
AAP-7538 docathon Modify Operations guide section 3 content

ariordan-redhat:
Part of docathon to refactor Installation guide

AAP-7538: Affects section 3 of Operations guide (`/titles/aap-operations-guide/`)

Close this PR if it's out of scope for the docathon: There were no structural/build issues to tackle in this section of the Operations Guide, so I just tidied up some asciidoc in the modules.

### DIFF
--- a/downstream/modules/platform/proc-control-data-collection.adoc
+++ b/downstream/modules/platform/proc-control-data-collection.adoc
@@ -13,5 +13,5 @@ You can control how {ControllerName} collects data by setting your participation
 ** *Off*: Prevents any data collection.
 ** *Anonymous*: Enables data collection without your specific user data.
 ** *Detailed*: Enables data collection including your specific user data.
-. Click *Save* to apply the settings or *Cancel* to abandon the changes.
+. Click btn:[Save] to apply the settings or btn:[Cancel] to discard the changes.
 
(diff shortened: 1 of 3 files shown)


downstream/modules/platform/proc-control-data-collection.adoc
```diff
@@ -1,20 +1,17 @@
...
 
-. Log in to your {ControllerName}
-
-. Navgate to *Settings* -> *User Interface*
-. Select the desired level of data collection from the User Analytics Tracking State drop-down list:
-.. *Off*: Prevents any data collection.
-.. *Anonymous*: Enables data collection without your specific user data.
-.. *Detailed*: Enables data collection including your specific user data.
+. Log in to your {ControllerName}.
+. Navigate to *Settings > User Interface*.
```
ariordan-redhat, line 11: Used *Settings > User Interface* here because the `menu:` asciidoc tag renders the menu items as item1 -> item2.
Which is better from an accessibility viewpoint?

dcdacosta: @ariordan-redhat From an accessibility standpoint, it's better to use the menu:item1[item2]. The rendering is something I've been trying to push DXP to fix but it keeps getting deferred. I wonder if we should discuss this again in the next team meeting?

PR #696, opened 2022-12-13
Corrected table (#695)

ianf77:
Added extra column and removed entries that are included in general system requirements

Correct table in Planning Guide

https://issues.redhat.com/browse/AAP-7672


ariordan-redhat (approved):
LGTM

PR #703, opened 2022-12-14
AAP-4995 - Add AWX deployment to list of migration options

dcdacosta:
This PR addresses the changes required by https://issues.redhat.com/browse/AAP-4995.

Change includes the following:

Added bullet to Chapter 6 to show that the procedures cover migration of AWX --> Openshift AAP Operator as well as Tower and VM deployments:

- An AWX instance

### DIFF
--- a/downstream/assemblies/platform/assembly-aap-migration.adoc
+++ b/downstream/assemblies/platform/assembly-aap-migration.adoc
@@ -10,10 +10,11 @@ ifdef::context[:parent-context: {context}]
 
 Migrating your {PlatformName} deployment to the {OperatorPlatform} allows you to take advantage of the benefits provided by a Kubernetes native operator, including simplified upgrades and full lifecycle support for your {PlatformName} deployments.
 
-This guide allows you to migrate:
+Use these procedures to migrate any of the following deployments to the {OperatorPlatform}:
 
 * A VM-based installation of {ControllerName} or {HubName}, or
 * An Openshift instance of Ansible Tower 3.8.6 ({PlatformNameShort} 1.2)
+* An AWX instance of Ansible Tower 3.8.6 ({PlatformNameShort} 1.2)
 
 
 include::platform/con-aap-migration-considerations.adoc[leveloffset=+1]
--- a/downstream/assemblies/platform/assembly-aap-migration.adoc
+++ b/downstream/assemblies/platform/assembly-aap-migration.adoc
@@ -12,8 +12,8 @@ Migrating your {PlatformName} deployment to the {OperatorPlatform} allows you to
 
 Use these procedures to migrate any of the following deployments to the {OperatorPlatform}:
 
-* A VM-based installation of {ControllerName} or {HubName}, or
-* An Openshift instance of Ansible Tower 3.8.6 ({PlatformNameShort} 1.2)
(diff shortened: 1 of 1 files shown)


downstream/assemblies/platform/assembly-aap-migration.adoc
```diff
@@ -10,10 +10,11 @@ ifdef::context[:parent-context: {context}]
...
 Migrating your {PlatformName} deployment to the {OperatorPlatform} allows you to take advantage of the benefits provided by a Kubernetes native operator, including simplified upgrades and full lifecycle support for your {PlatformName} deployments.
 
-This guide allows you to migrate:
+Use these procedures to migrate any of the following deployments to the {OperatorPlatform}:
 
-* A VM-based installation of {ControllerName} or {HubName}, or
-* An Openshift instance of Ansible Tower 3.8.6 ({PlatformNameShort} 1.2)
+* A VM-based installation of {ControllerName} or {HubName},
+* An Openshift instance of Ansible Tower 3.8.6 ({PlatformNameShort} 1.2), or
+* An AWX instance of Ansible Tower 3.8.6 ({PlatformNameShort} 1.2)
```
rooftopcellist, line 17: * An AWX instance of Ansible Tower 3.8.6 ({PlatformNameShort} 1.2)
AWX is the upstream project, Ansible Tower is the downstream product.  

So we'd want something more like:
* An AWX instance

Also, there may be cases where db migrations got merged upstream that were removed before downstream releases.  This would be rare, but possible.  So for this reason, I can't 100% gaurantee that this migration path will always work.  But there isn't much we can do about that.  

In cases like that, the TAM or Support engineer involved could help them by upgrading AWX and trying the migration again.

dcdacosta: @rooftopcellist 

I have updated the bullet but opted not to note the potential for AWX migration issues since, as you say, it's rare.

downstream/assemblies/platform/assembly-aap-migration.adoc
```diff
@@ -10,11 +10,11 @@ ifdef::context[:parent-context: {context}]
...
 Migrating your {PlatformName} deployment to the {OperatorPlatform} allows you to take advantage of the benefits provided by a Kubernetes native operator, including simplified upgrades and full lifecycle support for your {PlatformName} deployments.
 
-This guide allows you to migrate:
-
-* A VM-based installation of {ControllerName} or {HubName}, or
-* An Openshift instance of Ansible Tower 3.8.6 ({PlatformNameShort} 1.2)
+Use these procedures to migrate any of the following deployments to the {OperatorPlatform}:
 
+* A VM-based installation of {ControllerName} or {HubName},
+* An Openshift instance of Ansible Tower 3.8.6 ({PlatformNameShort} 1.2), or
```
ianf77, line 16: Is it still Ansible Tower in this version?

ianf77 (commented):
LGTM. One query.

downstream/assemblies/platform/assembly-aap-migration.adoc
dcdacosta: @ianf77 Yes, because this information is intended to help users migrate from old deployments to the new AAP Operator.

PR #757, opened 2023-01-19
AAP-8071 Wrong database guidelines have been updated in  planning guide

rogrange:
Wrong database guidelines have been corrected in the Planning Guide.

### DIFF
--- a/downstream/modules/platform/ref-guidelines-hosts-groups.adoc
+++ b/downstream/modules/platform/ref-guidelines-hosts-groups.adoc
@@ -4,7 +4,7 @@
 
 .Databases
 * When using an external database, ensure the `[database]` sections of your inventory file are properly set up.
-* The database server should be on the same network or in the same datacenter as the {ControllerNameStart} server for performance reasons.
+* For the best performance, ensure that your database server and the {ControllerNameStart} server are in the same network or datacenter.
 
 .{HubNameStart}
 * Add {HubNameMain} information in the `[automationhub]` group
@@ -20,7 +20,7 @@ If you use one value in `[database]` and both {Controllername} and {HubNameMain}
 ====
 
 .{ControllerNameStart}
-* {ControllerNameStart} will not configure replication or failover for the database that it uses, although automation controller should work with any replication that you have. 
+* {ControllerNameStart} will not configure replication or failover for the database that it uses.
 {ControllerName} works with any replication that you have.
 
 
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-guidelines-hosts-groups.adoc
```diff
@@ -4,7 +4,7 @@
 
 .Databases
 * When using an external database, ensure the `[database]` sections of your inventory file are properly set up.
-* To improve performance, do not place the database server and the automation controller server on the same network.
+* The database server should be on the same network or in the same datacenter as the {ControllerNameStart} server for performance reasons.
```
jself-sudoku, line 7: Please don't shoot me, Robert! I'm looking at the use of the word "should" knowing that we are supposed to avoid the use of it, if at all possible. Sorry, it jumped out at me. 
```suggestion
*For the best performance, ensure that your database server and the {ControllerNameStart} server are in the same network or datacenter.
```
Here's what the IBM styles and standards say under Word usage: 
"Use with caution should   verb
Whenever possible, rewrite in a more direct way. Do not use should to refer to an action that must be performed. For example, when parameters must be reset, write “Reset the parameters” and not “You should reset the parameters”."

downstream/modules/platform/ref-guidelines-hosts-groups.adoc
```diff
@@ -20,7 +20,7 @@ If you use one value in `[database]` and both {Controllername} and {HubNameMain}
 ====
 
 .{ControllerNameStart}
-* {ControllerNameStart} does not configure replication or failover for the database that it uses. 
+* {ControllerNameStart} will not configure replication or failover for the database that it uses, although automation controller should work with any replication that you have. 
 {ControllerName} works with any replication that you have.
```
jself-sudoku, line 24: Did you mean to remove the final sentence here since you pretty much said the same thing at the end of the previous sentence above? The only other thing I can say is that typically we try to avoid using future tense and use present tense words like "does" because it presents the message in a more direct way. I'll leave it up to you to whether you want to keep it with "does" instead of "will". 

For reference about future tense, here's a link: https://www.ibm.com/docs/en/ibm-style?topic=grammar-verbs#tense. It basically says write in present tense as much as possible, except where you cannot avoid it.

downstream/modules/platform/ref-guidelines-hosts-groups.adoc
```diff
@@ -20,8 +20,8 @@ If you use one value in `[database]` and both {Controllername} and {HubNameMain}
 ====
 
 .{ControllerNameStart}
-* {ControllerNameStart} does not configure replication or failover for the database that it uses. 
-{ControllerName} works with any replication that you have.
+* {ControllerNameStart} will not configure replication or failover for the database that it uses.
```
jself-sudoku, line 23: Hey. Have you considered returning the future tense (will) to present tense (does), which is preferred unless it can't be done?

jself-sudoku:
LGTM

PR #790, opened 2023-01-31
AAP-2081 Catalog Fix image syntax in catalog guides

ariordan-redhat:
Modify paths to images in catalog guides.
Part of AAP-2081

Affects
`titles/catalog/itsm-integration`


dcdacosta (approved):
LGTM! Downloaded this PR locally and built using bccutil and the affected titles run without error.

PR #791, opened 2023-01-31
Corrections to OCP Performance doc

ianf77:
Small changes to document based on SME's edits

Corrections to Performance Considerations document

https://issues.redhat.com/browse/AAP-8737

### DIFF
--- a/downstream/modules/platform/ref-container-resource-requirements.adoc
+++ b/downstream/modules/platform/ref-container-resource-requirements.adoc
@@ -24,7 +24,9 @@ See also xref:ref-schedule-jobs-worker-nodes[Jobs scheduled on the worker nodes]
 | `redis_resource_requirements` | Redis control plane container resource requirements | requests: {CPU:100m, memory: 128Mi}
 |====
 
-An example of customization could be:
+Because the use of `topology_spread_constraints` to maximally spread control nodes onto separate underlying kubernetes worker nodes is also recommended, a reasonable set of requests and limits would be limits whose sum is equal to the actual resources on the node. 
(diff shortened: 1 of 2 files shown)


ianf77:
kdelee@redhat.com Can you review these changes please?

downstream/modules/platform/ref-container-resource-requirements.adoc
```diff
@@ -21,6 +21,7 @@ See also xref:ref-schedule-jobs-worker-nodes[Jobs scheduled on the worker nodes]
 | `web_resource_requirements` | Web container resource requirements | requests: {CPU: 100m, memory: 128Mi}
 | `task_resource_requirements` | Task container resource requirements | requests: {CPU: 100m, memory: 128Mi}
 | `ee_resource_requirements` | EE control plane container resource requirements | requests: {CPU: 100m, memory: 128Mi}
+| `redis_resource_requirements` | Redis control plane container resource requirements | requests: {CPU:100m, memory: 128Mi}
 |====
 
 An example of customization could be:
```
kdelee, line 27:
```suggestion
Because we also recommend the use of topology_spread_constraints to maximally spread control nodes onto separate underlying kubernetes worker nodes, a reasonable set of requests and limits would be limits whose sum equal the actual resources on the node. If only limits are set, then the request is automatically set to be equal to the limit. But because we want to allow some variablity of resource usage between the containers in the control pod, you can set requests to a lower amount, for example to 25% of the resources available on the node. An example of container customization for a cluster where the worker nodes have 4 CPUs and 16 GB of RAM could be:
```

kdelee (approved):
left one suggestion

lmalivert:
@ianf77 LGTM. Merging.

PR #819, opened 2023-02-06
Added information on builds

ianf77:
Added text to describe effect of not specifying EE_BASE_IMAGE and not using the -t switch when building.

ansible builder docs EE_BASE_IMAGE implications

https://issues.redhat.com/browse/AAP-419


jself-sudoku:
LGTM

PR #823, opened 2023-02-07
Correct statement about database location

EMcWhinn:
Inserted correct text from Jira

Note in documentation about database and automation controller location is incorrect

https://issues.redhat.com/browse/AAP-8783


ianf77 (approved):
LGTM

PR #839, opened 2023-02-09
Added default values of image.

ianf77:
Updated table with default values (normal and with RH RPM build.

Correct default value for EE_XXXX_IMAGE

https://issues.redhat.com/browse/AAP-8980

### DIFF
--- a/downstream/modules/builder/ref-build-args-base-image.adoc
+++ b/downstream/modules/builder/ref-build-args-base-image.adoc
@@ -21,7 +21,7 @@ Note that, if you are using a Red Hat ansible-builder RPM, the default image bec
 | `EE_BUILDER_IMAGE`
 | Specifies the intermediate builder image used for Python dependency collection and compilation; must contain a matching Python version with EE_BASE_IMAGE and have ansible-builder installed.
 
-If no `EE_BASE_IMAGE` is specified, the default image `quay.io/ansible/ansible-builder:latest` is used.
+If no `EE_BUILDER_IMAGE` is specified, the default image `quay.io/ansible/ansible-builder:latest` is used.
 
 Note that, if you are using a Red Hat ansible-builder RPM, the default image becomes `registry.redhat.io/ansible-automation-platform-23/ansible-builder-rhel8:latest`.
 |===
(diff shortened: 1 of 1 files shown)


simaishi:
@ianf77 For downstream documentation, do we actually document multiple ways to install ansible-builder, not just RPM?

ianf77:
@simaishi I don't know. Who would be able to supply this information?

simaishi:
https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.3/html/creating_and_consuming_execution_environments

Is this the only guide that talks about ansible-builder?

Regardless, for downstream, the only 'supported' way should be RPM. So I'm not actually sure if we really need to document quay at all... Please check with Phil.

ianf77:
@simaishi 
According to Phil Griffiths quay is upstream only, downstream documents shouldn't mention it. How do you want to proceed here? Should the original statement be reworded to imply that RPM is the only supported installation method and that the value used with RPM is the default value?

simaishi:
@ianf77 I think we can just add defalut value to existing explanation:

| `EE_BASE_IMAGE`
| Specifies the parent image for the automation execution environment, enabling a new image to be built that is based off of an already-existing image. This is typically a supported execution environment base image like ee-minimal or ee-supported, but it can also be an execution environment image that you've created previously and want to customize further. **The default value is 'registry.redhat.io/xxxxx'**.

I don't think we need to mention RPM or not at all, since that's the only supported way and that's the instruction in this guide.

ianf77:
Corrections made.

simaishi:
@ianf77 Just realized 'main' branch is for AAP 2.4, not 2.3.  So the path to the image is: `registry.redhat.io/ansible-automation-platform-24/xxx`.  Looks good otherwise.

ianf77:
@simaishi 
The docs branch for 2.4 hasn't been created yet. It'll be created from main, but I can add an attribute which put the current version into the image.

ianf77:
Actually we don't have an attribute which would replace 23 or 24. can I suggest <latest version> or <aap_version>?

simaishi:
@ianf77 I'll leave it up to you.

ariordan-redhat (approved):
LGTM

PR #855, opened 2023-02-14
2.3 Added default values of image. (#839)

ianf77:
* Added default and rpm build values of image.

Updated table with default values (normal and with RH RPM build.

Correct default value for EE_XXXX_IMAGE

Affects `titles/builder/`

https://issues.redhat.com/browse/AAP-8980


ariordan-redhat (approved):
LGTM

PR #761, opened 2023-01-20
AAP-4208 updated clickable elements

emurtoug:
Per [AAP-4208](https://issues.redhat.com/browse/AAP-4208), updated clickable elements to reflect proper style: 

Instead of Click **Ok** use Click btn:[Ok].

PR #859, opened 2023-02-14
2.3 AAP-4208 updated clickable elements (#761)

ariordan-redhat:
Backports #761  to 2.3
* AAP-4208 updated button formating


dcdacosta (approved):
LGTM

PR #863, opened 2023-02-16
Removing instance of PostgreSQL statement (#862)

EMcWhinn:
Removing incorrect PostgreSQL statement as AAP 2.x doesn't use SCL.

Remove incorrect PostgreSQL statement.

https://issues.redhat.com/browse/AAP-9320


ianf77 (approved):
LGTM

PR #868, opened 2023-02-20
AAP-2080 added note about 301 redirect

rogrange:
Added a note stating that all api urls must end with a trailing slash or the user will receive a 301 redirect.


dcdacosta (approved):
LGTM

PR #881, opened 2023-02-23
AAP-7549 added variables for load balancing and header usage.

rogrange:
AAP-7549 added variables for load balancing and header usage.


hherbly:
LGTM @rogrange.

PR #882, opened 2023-02-27
Added nginx_tls_protocols to appendix

ianf77:
Added new variable

Document new variable 'nginx_tls_protocols' for AAP 2.2+

https://issues.redhat.com/browse/AAP-9490


simaishi:
We use a different conf file name, so I think `Defines support for ssl_protocols in Nginx` would be better.

ariordan-redhat (approved):
LGTM

PR #894, opened 2023-03-03
Modify module IDs to fix docs build problems

ariordan-redhat:
Bad syntax in module IDs was causing a docs build to fail.

Affects `titles/hub/manage-containers`


jself-sudoku:
LGTM

PR #898, opened 2023-03-07
Created templates for future release notes

dcdacosta:
This PR contains templates for future Release Notes updates. Templates with instructions at the top of each file are stored in the ../downstream/snippets directory.

### DIFF
--- a/downstream/snippets/aap-rn-template.adoc
+++ b/downstream/snippets/aap-rn-template.adoc
@@ -1,0 +1,74 @@
+// For each release of AAP, make a copy of this file and rename it to aap-rn-xx.adoc where xx is the release number; for example, 24 for the 2.4 release.
+// Save the renamed copy of this file to the release-notes/topics directory topic files for the release notes reside.
+//Only include release note types that have updates for a given release. For example, if there are no Technology previews for the release, remove that section from this file.
+
+
+= Ansible Automation Platform <version>
+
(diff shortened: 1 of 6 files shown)


lmalivert:
LGTM @dcdacosta!

ianf77 (commented):
A general comment. Where would we indicate documentation changes? For example new documents for 2.4, or when the Install Guide was split and Planning and Operations guide created. Or the addition of the Performance Guide. We ought to announce it somewhere.

ariordan-redhat:
Would it make sense to move the release notes to the `downstream` folder, ie at the same level as `titles/`?  It might be out of scope for this work.

jself-sudoku:
LGTM

hherbly:
LGTM!

emurtoug:
Would this template apply also to Z-Stream release notes? I'm thinking here of [AAP-5422](https://issues.redhat.com/browse/AAP-5422).

dcdacosta:
> A general comment. Where would we indicate documentation changes? For example new documents for 2.4, or when the Install Guide was split and Planning and Operations guide created. Or the addition of the Performance Guide. We ought to announce it somewhere.

@ianf77 Good question; My original thinking was that this type of information could be added in a more ad-hoc fashion like I did for 2.3. I thought it might limit confusion by limiting the number of templates but maybe that's flawed and might cause doc change info to be missed. I'll go ahead and add a template.

dcdacosta:
> Would it make sense to move the release notes to the `downstream` folder, ie at the same level as `titles/`? It might be out of scope for this work.

@ariordan-redhat I'm not sure if there was a specific reason for keeping it in a standalone directory? In my other projects, the release notes information was contained along side the other docs. Making this change might also allow us to use our attributes in the RN!?

dcdacosta:
> Would this template apply also to Z-Stream release notes? I'm thinking here of [AAP-5422](https://issues.redhat.com/browse/AAP-5422).

@emurtoug I would say, for consistency sake, yes. You would just set the placeholder for version to the correct ZStream.

PR #916, opened 2023-03-16
AAP-4587 Add SSO variables to hub inventory tables

ariordan-redhat:
Part of AAP-4587 "connecting Automation Hub to an existing Red Hat SSO environment"

Adds SSO variables to the Hub inventory variables tables.
Affects `/titles/aap-installation-guide`

### DIFF
--- a/downstream/modules/platform/ref-hub-variables.adoc
+++ b/downstream/modules/platform/ref-hub-variables.adoc
@@ -212,14 +212,14 @@ Default = `Ansible Automation Platform`
 
 Used for {PlatformNameShort} managed and externally managed {RHSSO}.
 
-SSO admin username.
+SSO administration username.
 
 Default = `admin`
 | *`sso_console_admin_password`* | _Required_
 
 Used for {PlatformNameShort} managed and externally managed {RHSSO}.
 
-SSO admin password.
+SSO administration password.
 //| *`sso_console_keystore_file`* | Keystore file to install in SSO node.
 //
 //`/path/to/sso.jks`
@@ -264,7 +264,7 @@ This must be reachable from client machines.
 
 Used for {PlatformNameShort} managed and externally managed {RHSSO}.
 
-Set to `true` if the cert is to be validated during connection.
+Set to `true` if the certificate is to be validated during connection.
 
 Default = `true`
 
--- a/downstream/modules/platform/ref-hub-variables.adoc
+++ b/downstream/modules/platform/ref-hub-variables.adoc
@@ -183,23 +183,50 @@ Default = `TLSv1.2`.
 The path is on the Ansible management node. 
 It is used to encrypt certain fields in the database (such as credentials.) 
 If not specified, a new key will be generated.
-| *`sso_console_admin_password`* | SSO admin password.
-| *`sso_console_keystore_file`* | Keystore file to install in SSO node.
+| *`sso_console_admin_username`* | SSO admin username.
 
-`/path/to/sso.jks`
+Default = `admin`
+| *`sso_console_admin_password`* | SSO admin password.
+//| *`sso_console_keystore_file`* | Keystore file to install in SSO node.
+//
+//`/path/to/sso.jks`
 | *`sso_host`* | {CatalogNameStart} requires SSO and SSO administration credentials for
 authentication. 
 
 If SSO is not provided in the inventory for configuration, then you must use this variable to define the SSO host.
-| *`sso_keystore_password`* | Whether a keystore password is required for https enabled SSO.
-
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-hub-variables.adoc
```diff
@@ -200,6 +200,32 @@ This must be reachable from client machines.
...
+| *`sso_automation_platform_realm`* | 
+
+Default = `ansible-automation-platform`
+| *`sso_automation_platform_realm_displayname`* | Display name for the realm.
+
+Default = `Ansible Automation Platform`
+| *`sso_automation_platform_login_theme`* | Path to the theme files.
+
+Default = `ansible-automation-platform`
+| *`sso_http_port`* or *`sso_https_port`* | IP or routable hostname for SSO.
```
ianf77, line 226: I would make these separate entries.

ianf77 (commented):
One comment

downstream/modules/platform/ref-hub-variables.adoc
ariordan-redhat: @ianf77 Satoe isn't sure whether these variables are necessary. I'm commenting them out in case they do become necessary.

ianf77 (approved):
LGTM

downstream/modules/platform/ref-hub-variables.adoc
```diff
@@ -200,6 +200,32 @@ This must be reachable from client machines.
...
+| *`sso_keystore_name`* | Name of keystore for SSO.
+
+Default = `ansible-automation-platform`
+| *`sso_automation_platform_realm`* | The name of the realm in SSO.
+
+Default = `ansible-automation-platform`
+| *`sso_automation_platform_realm_displayname`* | Display name for the realm.
+
+Default = `Ansible Automation Platform`
+| *`sso_automation_platform_login_theme`* | Path to the theme files.
```
simaishi, line 223: If changing this var, customers need to provide their own theme files and this var needs to be path to the folder where theme files are located.

downstream/modules/platform/ref-hub-variables.adoc
```diff
@@ -183,23 +183,50 @@ Default = `TLSv1.2`.
...
 | *`sso_use_https`* | If Single Sign On uses https.
 
 Default = `true`
+| *`sso_ssl_validate_certs`* | Set to `true` if the cert is to be validated during connection.
+
+Default = `true`
+| *`sso_keystore_name`* | Name of keystore for SSO.
+
+Default = `ansible-automation-platform`
+| *`sso_keystore_password`* | Whether a keystore password is required for https enabled SSO.
```
simaishi, line 209: `sso_keystore_password` is where customers need to specify the password for keystore. This is required when using AAP managed SSO and https is enabled (which is the default)

simaishi:
Not modified in this PR, but `sso_host` description mentions catalog, should be hub.

ariordan-redhat:
@simaishi 
I updated the SSO variables so they contain required/optional  and they reference whether it's for AAP or external SSO.
So the info is all in one place, but people using the hub installation guide still need to refer to another place for the variables.

Should I take the SSO variables out of the Hub variables table ?

That way, they can remain in the Installation Guide and we can reuse the SSO variables module in the Hub installation guide - that is, we still would have a single source of truth but the customer wouldn't have to look in 2 places.

downstream/modules/platform/ref-hub-variables.adoc
```diff
@@ -183,21 +183,96 @@ Default = `TLSv1.2`.
...
+
+If `sso_redirect_host` is set, it is used by the application to connect to SSO for authentication. 
 
 This must be reachable from client machines.
-| *`sso_use_https`* | If Single Sign On uses https.
+| *`sso_ssl_validate_certs`* | _Optional_
+
+Used for {PlatformNameShort} managed and externally managed {RHSSO}.
+
+Set to `true` if the cert is to be validated during connection.
```
ianf77, line 267: certificate

downstream/modules/platform/ref-hub-variables.adoc
```diff
@@ -183,21 +183,96 @@ Default = `TLSv1.2`.
...
+Used for {PlatformNameShort} managed and externally managed {RHSSO}.
+
+SSO admin username.
+
+Default = `admin`
+| *`sso_console_admin_password`* | _Required_
+
+Used for {PlatformNameShort} managed and externally managed {RHSSO}.
+
+SSO admin password.
```
ianf77, line 222: Is admin, 'administration' ?

PR #928, opened 2023-03-28
Moved 'routable_hostname' variable

ianf77:
Moved routable hostname from controller to general variables.

Variable routable_hostname should be moved to A1 general section rather than controller specific

https://issues.redhat.com/browse/AAP-10627


ariordan-redhat (approved):
LGTM

PR #940, opened 2023-04-03
AAP-10662 Update AAP version to 2.4

ariordan-redhat:
- Updates version in docinfo.xml from `2.3` to `2.4`
- Updates OpenShift version

**NOTE:**
- Does not update core version or path for AAP download.
- Does not update files whose version is older than `2.3`.
- Does not scrub `2.3` from the general docset


ianf77 (approved):
LGTM

PR #976, opened 2023-04-20
AAP-11388 Add sync script for AAP 2.4

ariordan-redhat:
Create a sync script in the `bin/` directory that will be deployed by the Jenkins server to sync the AAP 2.4 docs to the production server.


ianf77 (approved):
LGTM

PR #1021, opened 2023-05-16
AAP-10050 Delete release notes from downstream folder

ariordan-redhat:
Delete release notes from downstream folder.
Release notes are in `downstream/titles/release-notes`


ianf77 (approved):
LGTM

PR #1022, opened 2023-05-16
AAP-2079 Add symlink to images dir

ariordan-redhat:
AAP-2079 Add symlink to images dir for Installation guide


dcdacosta (approved):
LGTM and builds locally without error.

PR #1023, opened 2023-05-16
Fix note syntax

ariordan-redhat:
Fixed asciidoc note syntax: it was generating a warning during the build process.
Affects `titles/aap-installation-guide/`


hherbly (approved):
LGTM

PR #1027, opened 2023-05-18
AAP-12313: adds an underscore to an automation controller variable 

hherbly:
Fixes issue in [AAP-12313](https://issues.redhat.com/browse/AAP-12313). For more context, see this [Slack thread](https://redhat-internal.slack.com/archives/C035VBWU98D/p1684429969208019). 

This change will add an underscore to an automation controller variable.

### DIFF
--- a/downstream/modules/platform/ref-controller-variables.adoc
+++ b/downstream/modules/platform/ref-controller-variables.adoc
@@ -6,7 +6,7 @@
 |====
 | *Variable* | *Description* 
 | *`admin_password`* | The password for an administration user to access the UI upon install completion.
-| *`automationcontroller_main_url`* | For an alternative front end URL needed for SSO configuration with {CatalogName}, provide the URL.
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-controller-variables.adoc
```diff
@@ -6,7 +6,7 @@
 |====
 | *Variable* | *Description* 
 | *`admin_password`* | The password for an administration user to access the UI upon install completion.
-| *`automationcontroller_main_url`* | For an alternative front end URL needed for SSO configuration with {CatalogName}, provide the URL.
+| *`automation_controller_main_url`* | For an alternative front end URL needed for SSO configuration with {CatalogName}, provide the URL.
```
ianf77, line 9: If this is for 2.4 references to {CatalogNameStart} should be removed.

hherbly: Hi Ian, thank you for your review--I'll remove {CatalogName} in a separate PR. That way I can check for any other occurrences in 2.4. Would you be willing to squash and merge this change in the meantime?

PR #1032, opened 2023-05-24
AAP-11398 Archive unused assemblies and modules

ariordan-redhat:
AAP-11398 Archive unused assemblies and modules for AAP 2.4

PR #1035, opened 2023-05-25
2.4 AAP-11398 Archive unused assemblies and modules (#1032)

ariordan-redhat:
AAP-11398 Archive unused assemblies and modules for 2.4


dcdacosta (approved):
LGTM!

PR #1036, opened 2023-05-26
AAP-11480 updated Getting Started to include instructions for PAH

emurtoug:
[AAP-11480](https://issues.redhat.com/browse/AAP-11480)

Updated [Chapter 1: Creating the Red Hat automation hub API Token](https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/1.2/html/getting_started_with_red_hat_ansible_automation_hub/proc-create-api-token) to include a subsection 1.2 for creating the API token in PAH. Procedure differs slightly in PAH from procedure in automation hub.

### DIFF
--- a/downstream/modules/automation-hub/con-create-api-token.adoc
+++ b/downstream/modules/automation-hub/con-create-api-token.adoc
@@ -1,0 +1,10 @@
+[id="con-create-api-token"]
+= Creating the Red Hat automation hub API token
+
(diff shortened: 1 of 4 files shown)


ianf77 (approved):
I can't see anything immediate that might stop the attributes from rendering. The one thing it might have been is present (that's the :experimental: label in master.adoc.). The rest of the doc builds with attributes, so it's something about the new files. But if some render and not others, then I'm just as confused as you. Approving anyway!

PR #1038, opened 2023-05-26
[2.3] AAP-8366: Recommend using fully updated nodes to install AAP (#1029)

lmalivert:
Jira link: [AAP-8366](https://issues.redhat.com/browse/AAP-8366)

Added prerequisite and note stating that users should use fully updated nodes prior to installing AAP.


emurtoug:
LGTM

PR #1050, opened 2023-06-02
AAP-10975C New chapter on EDA installation w/ AAP operator on OCP

jself-sudoku:
Created new assembly and module for instructions on installing EDA on AAP operator on OpenShift Container Platform. Also added new assembly to master. adoc.

### DIFF
--- a/downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
+++ b/downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
@@ -10,9 +10,11 @@
 
 .Procedure
 
-. Navigate to menu:Operators[Installed Operators].
+. Select menu:Operators[Installed Operators].
 
-. Create a new Event-Driven Ansible custom resource. 
+. Locate and select your installation of {PlatformNameShort}.
+
+. Under *Provided APIs*, locate the {EDAName} modal and click *Create instance*. 
 +
 This takes you to the Form View to customize your installation.
 
@@ -26,17 +28,23 @@ This is the only required customization, but you can customize other options usi
 ====
 
 . Click btn:[Create].
-This deploys {EDAcontroller} in the namespace you specified. After a couple minutes when the installation is marked as *Successful*, you can reach the UI. 
+This deploys {EDAcontroller} in the namespace you specified. 
++
+After a couple minutes when the installation is marked as *Successful*, you can find the URL for the {EDAName} UI on the *Routes* page in the Openshift UI. 
 
-. Click *Routes* to find the new Route URL that has been created for you.
+. From the navigation panel, select menu:Networking[Routes] to find the new Route URL that has been created for you. 
++
+Routes are listed according to the name of your custom resource.
 
 . Click the new URL to navigate to {EDAName} in the browser.
 
-. On the *Secrets* page in Openshift, locate the Admin Password k8s secret that was created for you, unless you specified a custom one. 
+. From the navigation panel, select menu:Workloads[Secrets] and locate the Admin Password k8s secret that was created for you, unless you specified a custom one.
++
+Secrets are listed according to the name of your custom resource and appended with `-admin-password.`
 +
 [NOTE]
 ====
-You can use the password value in the secret to log in to the {EDAcontroller} UI. The default user is `Admin`.
+You can use the password value in the secret to log in to the {EDAcontroller} UI. The default user is `admin`.
 ====
 
 
--- a/downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
+++ b/downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
@@ -16,7 +16,7 @@
 +
 This takes you to the Form View to customize your installation.
 
-. Specify your Controller URL. 
+. Specify your controller URL. 
 +
 If you deployed automation controller in Openshift as well, you can find the URL on the Routes page.
 +
--- a/downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
+++ b/downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
@@ -33,10 +33,10 @@ This deploys {EDAcontroller} in the namespace you specified. After a couple minu
 . Click the new URL to navigate to {EDAName} in the browser.
 
 . On the *Secrets* page in Openshift, locate the Admin Password k8s secret that was created for you, unless you specified a custom one. 
-
++
 [NOTE]
 ====
-The password value in that secret can be used to log in to the {EDAcontroller} UI.  The default user is `Admin`.
+You can use the password value in the secret to log in to the {EDAcontroller} UI. The default user is `Admin`.
 ====
 
 
--- a/downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
+++ b/downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
@@ -1,0 +1,45 @@
+
+[id="proc-deploy-eda-controller-with-aap-operator-ocp_{context}"]
+
+
+
+.Prerequisites
+
+* You have installed {OperatorPlatform} on {OCPShort}.
+* You have installed and configured {ControllerName}.
+
+.Procedure
+
+. Navigate to menu:Operators[Installed Operators].
+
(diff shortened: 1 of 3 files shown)


downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
```diff
@@ -0,0 +1,45 @@
...
++
+This takes you to the Form View to customize your installation.
+
+. Specify your controller URL. 
++
+If you deployed automation controller in Openshift as well, you can find the URL on the Routes page.
++
+[NOTE]
+====
+This is the only required customization, but you can customize other options using the UI form or directly in the YAML configuration tab, if desired.
```
rooftopcellist, line 25: We should find a way to document the "other options" eventually.  I think that might need a two-fold approach:
* There is a way for us to provide "in-line" documentation in the operator code that will show up in a help bar in the Openshift UI.  That way the docs are right where they will be used.
* I think for the RH docs, we should probably add a separate section or page for "Customizing your EDA Deployment" or something like that.  Then we could add things like that there.  But it might be better to use that space for capturing the "use cases/scenarios" that those advanced configuration setting would be used.  

I think that's outside the scope of this PR though, just dropping ideas here.

rooftopcellist:
This can be a separate PR, but we probably need to also add an external database section, with pretty much exactly the same content Automation Controller's section 3.3 has in it:
* [https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.3[…]_openshift_container_platform/installing-controller-operator](https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.3/html/deploying_the_red_hat_ansible_automation_platform_operator_on_openshift_container_platform/installing-controller-operator#proc-operator-external-db-controller)

rooftopcellist (approved):

downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
jself-sudoku: Thanks for your suggestions, @rooftopcellist. Per discussion, we will create a jira to address the customizing EDA options.

jself-sudoku:
@rooftopcellist - We will address both of your comments in separate jiras post-GA, per our discussion in Slack. Thank you for your review!

downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
```diff
@@ -0,0 +1,45 @@
...
+====
+
+. Click btn:[Create].
+This deploys {EDAcontroller} in the namespace you specified. After a couple minutes when the installation is marked as *Successful*, you can reach the UI.
```
dcdacosta, line 29: I'm confused by the phrase, "you can reach the UI".  Are they taken outside of the OpenShift UI to some other UI?

downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+.Prerequisites
+
+* You have installed {OperatorPlatform} on {OCPShort}.
+* You have installed and configured {ControllerName}.
+
+.Procedure
+
+. Navigate to menu:Operators[Installed Operators].
+
```
dcdacosta, line 14: I was testing this out in my test instance of AAP on OpenShift and I think there's more to this procedure:
1. Select Operators->Installed Operators. 
2. Locate and select your installation of Ansible Automation Platform.
3. Locate the EDA modal under Provided APIs and click Create instance.
This takes you to the Form View to customize your installation.

downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
```diff
@@ -0,0 +1,45 @@
...
++
+[NOTE]
+====
+This is the only required customization, but you can customize other options using the UI form or directly in the YAML configuration tab, if desired.
+====
+
+. Click btn:[Create].
+This deploys {EDAcontroller} in the namespace you specified. After a couple minutes when the installation is marked as *Successful*, you can reach the UI. 
+
+. Click *Routes* to find the new Route URL that has been created for you.
```
dcdacosta, line 31: This should be:
From the navigation panel, select menu:Networking[Routes] to find the new Route URL that has been created for you. Routes are listed according to the name of your custom resource.

downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+. Click the new URL to navigate to {EDAName} in the browser.
+
+. On the *Secrets* page in Openshift, locate the Admin Password k8s secret that was created for you, unless you specified a custom one.
```
dcdacosta, line 35: This should be:
From the navigation panel, select menu:Workloads[Secrets] and locate the Admin Password k8s secret that was created for you, unless you specified a custom one. Secrets are listed according to the name of your custom resource and appended with `-admin-password`.

downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+. Click *Routes* to find the new Route URL that has been created for you.
+
+. Click the new URL to navigate to {EDAName} in the browser.
+
+. On the *Secrets* page in Openshift, locate the Admin Password k8s secret that was created for you, unless you specified a custom one. 
++
+[NOTE]
+====
+You can use the password value in the secret to log in to the {EDAcontroller} UI. The default user is `Admin`.
```
dcdacosta, line 39: I think it's admin lowercase.

dcdacosta (changes requested):
Hey @jself-sudoku , I went through this procedure in my Openshift test environment and updated the procedures accordingly. Let me know if you want me to demo this for you so you can see how it works. Don't hesitate to let me know if you have questions.

downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
jself-sudoku: @dcdacosta  - for the first step, do we still use the accessibility method for this: "Select menu:Operators[installed Operators]? I'll do it that way for now and you can let me know if it's correct when you peer review this second attempt.

jself-sudoku: Also, @dcdacosta, should the step "Create a new Event-Driven Ansible custom resource" be removed now that you've added the first 3 steps above?

downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
rooftopcellist: I can add a little more context here.  When you deploy EDA using the AAP Operator on Openshift, the following is happening behind the scenes:
* containers are started inside the Openshift cluster which run the services needed to run the EDA application
* The service that provides the User Interface for EDA is not accessible via the browser until it is "exposed" publicly using Service and Route resources (which are automatically set up  by the operator when deploying EDA.
* Openshift Routes do some DNS magic that creates a URL for the EDA UI that is publically available.  So once the containers are all running and the route is created (all handled by the operator), the user should be able to access/reach the URL in their browser. 

tl;dr, when the deployment is complete, the URL for the EDA UI (found on the Route page in the Openshift UI), will be available/reacheable/accessible in the user navigates to it in their browser.

downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
dcdacosta: @jself-sudoku, I think you can delete it since that is what you're doing in this procedure. If you wanted to keep it, I would add it to the intro of the procedure as "Use this procedure to create a new Event-Driven Ansible custom resource".

rooftopcellist (approved):

downstream/modules/platform/proc-deploy-eda-controller-with-aap-operator-ocp.adoc
rooftopcellist: This is correct, it is lowercase.

dcdacosta (approved):
LGTM!

PR #1052, opened 2023-06-05
AAP-12753A Deleted extra EDA attributes added previously.

jself-sudoku:
Just deleted attributes that were previously added prior to confirming with the SMEs for EDA.


EMcWhinn:
LGTM

PR #1058, opened 2023-06-07
[2.2] AAP-8366: Use fully updated nodes to install AAP

lmalivert:
Jira link: [AAP-8366](https://issues.redhat.com/browse/AAP-8366)

Added prerequisite and note stating that users should use fully updated nodes prior to installing AAP.

See [PR#1029](https://github.com/RedHatInsights/red-hat-ansible-automation-platform-documentation/pull/1029) where the same information was added to 2.3 and 2.4.


hherbly (approved):
LGTM! My only comment is, should it be the *most* recent version of your RHEL nodes, or any recent version? If that doesn't matter, please disregard the comment.

lmalivert:
> LGTM! My only comment is, should it be the _most_ recent version of your RHEL nodes, or any recent version? If that doesn't matter, please disregard the comment.

Hey @hherbly! Thanks for the review. 
This has already received a technical review by the SME. He revised it to phrase it that way so it's my assumption the word "most" isn't doesn't make much of a difference.

PR #1075, opened 2023-06-15
Modified title fields to remove attribute and match to fix build (#1073)

dcdacosta:
edafix - Modified title fields to remove attribute and match to fix build issue


dcdacosta:
Writer made this change in the original PR, so I'm closing this PR as a duplicate.

PR #1084, opened 2023-06-19
Updated Getting Started guide and corrected errors on previous branch…

emurtoug:
Previous branch AAP-11480 (commit 15d554e) (PR [PR 1036](https://github.com/RedHatInsights/red-hat-ansible-automation-platform-documentation/pull/1036)) contained errors.

Created new branch AAP-11480B to correct errors.

Changes tracked in PR 1036 were merged to main but not backported. 

Only this current PR will be backported (to 2.3 and 2.4) after merge to main.

### DIFF
--- a/downstream/modules/automation-hub/con-create-api-token.adoc
+++ b/downstream/modules/automation-hub/con-create-api-token.adoc
@@ -3,7 +3,7 @@
 
 Before you can interact with automation hub by uploading or downloading collections, you must create an API token. The automation hub API token authenticates your `ansible-galaxy` client to the Red Hat automation hub server.
 
-You can create an API token using automation hub *Token management* in Automation hub or private automation hub (PAH).
+You can create an API token using automation hub *Token management* in automation hub or *API Token Management* in private automation hub (PAH).
 
 include::proc-create-api-token.adoc[leveloffset=+1]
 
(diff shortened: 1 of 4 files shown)


downstream/modules/automation-hub/con-create-api-token.adoc
```diff
@@ -3,7 +3,7 @@
 
 Before you can interact with automation hub by uploading or downloading collections, you must create an API token. The automation hub API token authenticates your `ansible-galaxy` client to the Red Hat automation hub server.
```
EMcWhinn, line 4: Should we use attribute for "automation hub"?

emurtoug: Hey @EMcWhinn! Thanks for this note. For some reason, the attribute was not rendering correctly in this document. It was repeatedly capitalising 'Automation Hub', although our style guide has it as 'automation hub'. Neither Ian nor I could figure out why that was happening, but I decided just to spell out automation hub instead of using the attribute in these cases. Post-2.4, I'll revisit this, and if I can resolve the attribute capitalisation error, I'll change these to attributes.

EMcWhinn: Hey @emurtoug, I think this happened to me recently as well now that you mention, so I had to spell it out too. All looks good so!

PR #1088, opened 2023-06-21
Fix tag in docinfo.xml

ariordan-redhat:
Fix subtitle tags in `docinfo.xml` for the following docs:

- `./upgrade/docinfo.xml`
- `./aap-operator-installation/docinfo.xml`
- `./security-guide/docinfo.xml`


emurtoug (approved):
LGTM

PR #1090, opened 2023-06-21
Created new content for synching images.

rogrange:
New content for synching images from a container repository.


hherbly:
LGTM!

PR #1099, opened 2023-06-22
Add section on controller max_connections demands

AlanCoding:
We have been doing work auditing and policing the database connections used by AAP controller / AWX. We want to document the maximum connection use envelope, and maybe this is a good place to do that. I just want to get things started.

I would like to link the other doc

https://github.com/RedHatInsights/red-hat-ansible-automation-platform-documentation/blob/a32872666f76c8ed3e1f0a5259d7d1b0c66ea82a/downstream/modules/platform/con-alternative-capacity-limits.adoc#L13

Since that's one of the 2 main relief values offered here, if the app is using too many connections, essentially.

### DIFF
--- a/downstream/modules/platform/ref-postgresql-requirements.adoc
+++ b/downstream/modules/platform/ref-postgresql-requirements.adoc
@@ -82,5 +82,5 @@ Adding up all the components comes out to 506 for this example cluster.
 Practically, this means that the max_connections should be set to something higher than this.
 Additional connections should be added to account for other platform components.
 
-This calculation is most sensitive to the number of forks per node. Database connections are opened at the start of and end of jobs. Environments where bursts of many jobs start at once will be most likely to reach the theoretical max number of open database connections.
+This calculation is most sensitive to the number of forks per node. Database connections are briefly opened at the start of and end of jobs. Environments where bursts of many jobs start at once will be most likely to reach the theoretical max number of open database connections.
 The max number of jobs that would be started concurrently can be adjusted by modifying the effective capacity of the instances. This can be done with the SYSTEM_TASK_ABS_MEM setting, the capacity adjustment on instances, or with instance groups max jobs or max forks.
--- a/downstream/modules/platform/ref-postgresql-requirements.adoc
+++ b/downstream/modules/platform/ref-postgresql-requirements.adoc
@@ -50,3 +50,37 @@ maintenance_work_mem == ansible_memtotal_mb*0.04
 
 .Additional resources
 For more detail on tuning your PostgreSQL server, see the link:https://wiki.postgresql.org/wiki/Main_Page[PostgreSQL documentation].
+
+|===
+
+.PostgreSQL Max Connections Setting
+
+For a more realistic method of determining a value of max_connections, a ballpark formula for {ControllerName} is outlined here.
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-postgresql-requirements.adoc
```diff
@@ -50,3 +50,37 @@ maintenance_work_mem == ansible_memtotal_mb*0.04
...
+  (16 connections / node) +
+  (4 connections / node)
+) + (5 connections)
+-----
+
+Adding up all the components comes out to 506 for this example cluster.
+Practically, this means that the max_connections should be set to something higher than this.
+Additional connections should be added to account for other platform components.
+
+This calculation is most sensitive to the number of forks per node.
```
kdelee, line 85:
```suggestion
This calculation is most sensitive to the number of forks per node. Database connections are opened at the start of and end of jobs. Environments where bursts of many jobs start at once will be most likely to reach the theoretical max number of open database connections.
```

downstream/modules/platform/ref-postgresql-requirements.adoc
```diff
@@ -50,3 +50,37 @@ maintenance_work_mem == ansible_memtotal_mb*0.04
...
+This can be adjusted with the SYSTEM_TASK_ABS_MEM setting, or with instance groups max jobs or max forks.
```
kdelee, line 86:
```suggestion
The max number of jobs that would be started concurrently can be adjusted by modifying the effective capacity of the instances. This can be done with the SYSTEM_TASK_ABS_MEM setting, the capacity adjustment on instances, or with instance groups max jobs or max forks.
```

downstream/modules/platform/ref-postgresql-requirements.adoc
```diff
@@ -50,3 +50,37 @@ maintenance_work_mem == ansible_memtotal_mb*0.04
...
+  (16 connections / node) +
+  (4 connections / node)
+) + (5 connections)
+-----
+
+Adding up all the components comes out to 506 for this example cluster.
+Practically, this means that the max_connections should be set to something higher than this.
+Additional connections should be added to account for other platform components.
+
+This calculation is most sensitive to the number of forks per node. Database connections are opened at the start of and end of jobs. Environments where bursts of many jobs start at once will be most likely to reach the theoretical max number of open database connections.
```
kdelee, line 85:
```suggestion
This calculation is most sensitive to the number of forks per node. Database connections are briefly opened at the start of and end of jobs. Environments where bursts of many jobs start at once will be most likely to reach the theoretical max number of open database connections.
```

kdelee (approved):

AlanCoding:
The one thing not said here is the source for `(132 forks / node)`. This effectively comes from RAM / memory amount with the ratio of memory per fork. That ratio is 100 (MB / fork) here:

https://github.com/ansible/awx/blob/devel/awx/main/utils/common.py#L858

But it can also be customized. Also, that feels like maybe a little bit low, and thus it should be customized. Anyway, the formula also has some powers-of-two conversions, and the 132 number came from experimentation. I can get close to that by hand calculations but it can be somewhat difficult.

AlanCoding:
Closing this for now, in order to indicate my intention to document the new installer setting.

After that, my plan is to migrate this content into the AWX docsite, which will eventually make it into the downstream docs here.

PR #1136, opened 2023-06-27
Corrected the link to the EDA controller user's guide

jself-sudoku:
Updated the link to the EDA controller's guide (still contain temporary link)


jself-sudoku:
In interest of GA preparedness, squashing and merging so that doc is correct for the release.

PR #1152, opened 2023-06-29
Changed icons in user guide

ianf77:
Changed format of icons to make them more visible

Icons in EDA user guide are barely visible.

https://issues.redhat.com/browse/AAP-13642

### DIFF
--- a/downstream/modules/eda/proc-eda-delete-rulebook-activations.adoc
+++ b/downstream/modules/eda/proc-eda-delete-rulebook-activations.adoc
@@ -2,7 +2,7 @@
 
 == Delete rulebook activations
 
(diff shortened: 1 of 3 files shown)


ianf77:
Hi Ed. I've changed the icons in this guide. We think that a  change to the Customer Portal has messed up the sizing we used for icons. This is another way of doing the same thing.
Just a question: In proc-eda-set-up-rulebook-activation is the variable '-e or --extra-vars' or are the variableS '-e' or '--extra-vars' ?

EMcWhinn:
> 

Hi Ian, I think this should be the latter '-e' or '--extra-vars', thanks for spotting it.

EMcWhinn:
LGTM

PR #1161, opened 2023-06-29
Use unicode characters for icons

ariordan-redhat:
Inline images that represented icons do not display correctly on the customer portal.
Changes to the portal can also introduce errors.
Use the Unicode characters recommended by the HCC to show inline icons.
Use an attribute to represent the icons.

Example of badly rendered "more actions" icon:

image


ianf77 (approved):
LGTM

dcdacosta (approved):
LGTM!

PR #1179, opened 2023-07-04
Replace acme.org with example.com

ianf77:
Mainly in inventory file examples

[DDF] acme.org is not owned by Red Hat, please use example.com or similar, explicitly reserved for documentation, domains

https://issues.redhat.com/browse/AAP-13632

### DIFF
--- a/downstream/modules/platform/ref-example-platform-ext-database-customer-provided.adoc
+++ b/downstream/modules/platform/ref-example-platform-ext-database-customer-provided.adoc
@@ -24,8 +24,8 @@ This example does not have a host under the database group. This indicates to th
 # hybrid2.example <- this will default to hybrid
 
 [automationcontroller]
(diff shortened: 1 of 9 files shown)


evgeni:
There are still two mentions of `acme.org` in the files, but I have no idea where those are rendered:

```
% git grep acme\.
downstream/archive/archived-modules/platform/ref-external-existing-database.adoc:tower.acme.org
downstream/modules/platform/ref-standalone-hub-ext-database-inventory.adoc:data.acme.org
```

Otherwise LGTM!

ianf77:
The one marked archive/ is in an archive folder as it's not used in any document. It can be ignored. Found the other file and fixing it now.

evgeni:
that works for me :)

thanks ian!

emurtoug (commented):
LGTM

PR #1181, opened 2023-07-04
Replace acme.org with example.com (#1179)

ianf77:
* Replace acme.org with example.com

Mainly in inventory file examples

[DDF] acme.org is not owned by Red Hat, please use example.com or similar, explicitly reserved for documentation, domains

https://issues.redhat.com/browse/AAP-13632

* Replace acme.org with example.com

Caught missing instance

[DDF] acme.org is not owned by Red Hat, please use example.com or similar, explicitly reserved for documentation, domains

https://issues.redhat.com/browse/AAP-13632


emurtoug (commented):
LGTM

PR #1198, opened 2023-07-13
Delete empty additional resources section

ariordan-redhat:
Delete empty additional resources section.
It was causing the module following it to be enclosed in a box.

image


ianf77 (approved):
LGTM

PR #1201, opened 2023-07-13
2.3 Delete empty additional resources section (#1198)

ariordan-redhat:
Backports #1198 to 2.3

Delete empty additional resources section.
It was causing the module following it to be enclosed in a box.

image


ianf77 (approved):
LGTM

PR #1226, opened 2023-07-18
2.4 Remove empty Additional Resources sections (#1203)

ariordan-redhat:
Empty additional resources sections are causing the assemblies or modules that follow them to be enclosed in a box.
Delete empty sections.


lmalivert:
LGTM @ariordan-redhat 👍🏾

PR #1235, opened 2023-07-21
Updating EDA controller user guide asciidoc

EMcWhinn:
Making necessary changes as per content audit

Updating EDA controller user guide asciidoc elements

https://issues.redhat.com/browse/AAP-13856

Affects `titles/eda-user-guide`


downstream/assemblies/eda/assembly-eda-projects.adoc
```diff
@@ -0,0 +1,9 @@
+[id="eda-projects"]
+
+= Projects
+
+[role="_abstract"]
+
+Projects are a logical collection of rulebooks.
+
+include::eda/proc-eda-set-up-new-project.adoc[leveloffset=-0]
```
dcdacosta, line 9: This should be [leveloffset=+1] which indicates that it is a subsection of this assembly. Then in the procedure module, the title should have only 1 = preceding the title text. I'll comment there as well. I believe nesting was previously managed with the = in the modules but now that we're using assemblies, the nesting is controlled by the leveloffset.

downstream/modules/eda/proc-eda-set-up-new-project.adoc
```diff
@@ -1,20 +1,20 @@
 [id="eda-set-up-new-project"]
 
-= Setting up a new project
+== Setting up a new project
```
dcdacosta, line 3: This should be = Setting up a new project and the extra = should be removed. See above for the explanation and don't hesitate to let me know if you need me to clarify further.

dcdacosta (approved):
Hi @EMcWhinn, Great job on this, I left just a few minor comments.

downstream/assemblies/eda/assembly-eda-projects.adoc
EMcWhinn: Done :)

dcdacosta (approved):
LGTM!

PR #1238, opened 2023-07-24
AAP-14310 updated network ports & protocols diagram

emurtoug:
[AAP-14310](https://issues.redhat.com/browse/AAP-14310)

Updated [Ch 5. Network Ports and Protocols](https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.4/html/red_hat_ansible_automation_platform_planning_guide/ref-network-ports-protocols_planning#doc-wrapper)

- Replaced architectural diagram with updated diagram that includes EDA


lmalivert:
@emurtoug Looks good to me 👍🏾 
This may be out of scope of this Jira issue, but I wonder if the tables section need to be revised and EDA information added?

emurtoug:
> @emurtoug Looks good to me 👍🏾 This may be out of scope of this Jira issue, but I wonder if the tables section need to be revised and EDA information added?

Thanks @lmalivert! Let's ask the WG about this.

PR #1257, opened 2023-07-27
Replace the word 'leverage'

ianf77:
Use 'use' instead, as per IBM style guide.

Change wording

https://issues.redhat.com/browse/AAP-14494


ariordan-redhat (approved):
Thank you!
LGTM

PR #1258, opened 2023-07-27
Replace the word 'leverage' (#1257)

ianf77:
Use 'use' instead, as per IBM style guide.
https://issues.redhat.com/browse/AAP-14494


ariordan-redhat (approved):
LGTM

PR #1270, opened 2023-07-31
Replaced attribute with Unicode icon

ianf77:
Testing showed that the unicode hex value worked best

Add missing icon

https://issues.redhat.com/browse/AAP-14546


ariordan-redhat (approved):
LGTM

PR #1272, opened 2023-07-31
Replace settings icon with unicode image

ianf77:
No attribute to replace so replaced it directly

Add missing icon

https://issues.redhat.com/browse/AAP-14546


ariordan-redhat (approved):
LGTM

PR #1299, opened 2023-08-11
Separate steps in procedure

ianf77:
Made one step into  two

[DDF] Running this step as listed on RHEL 9.2 running AAP 2.4 did not work. However, breaking it into two steps were I first

https://issues.redhat.com/browse/AAP-14866

### DIFF
--- a/downstream/modules/platform/proc-create-password-hashes.adoc
+++ b/downstream/modules/platform/proc-create-password-hashes.adoc
@@ -7,10 +7,15 @@
 +
 [literal, options="nowrap" subs="+quotes,attributes"]
 ----
-awx-manage shell_plus \
-from awx.main.utils import encrypt_value, get_encryption_key \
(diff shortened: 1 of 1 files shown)


msilmser (approved):

downstream/modules/platform/proc-create-password-hashes.adoc
```diff
@@ -7,10 +7,15 @@
 +
 [literal, options="nowrap" subs="+quotes,attributes"]
 ----
-awx-manage shell_plus \
-from awx.main.utils import encrypt_value, get_encryption_key \
-postgres_secret = encrypt_value('$POSTGRES_PASS') \
-print(postgres_secret)
+# awx-manage shell_plus
+----
+. Then run the following from the python prompt
```
EMcWhinn, line 12: Should a colon be included here? 

"Then run the following from the python prompt:"

EMcWhinn (approved):
LGTM

PR #1313, opened 2023-09-25
Document installer variable for max_connections

AlanCoding:
This is a follow-up change after @kdelee made a change to the installer that allows users to set this variable.

The content I reference in the administration guide will be https://github.com/RedHatInsights/red-hat-ansible-automation-platform-documentation/pull/1099, but first put upstream, which should eventually make it back here.


ariordan-redhat:
@AlanCoding 
We are no longer using this repo: we have migrated the docs to the aap-docs repo in the Ansible GitHub org:
https://github.com/ansible/aap-docs

